PR #40, opened 2019-10-24
Changed default for metrics with only value, metric, timestamp

srclosson:
Changing the default to be backwards compatible with the 1.3 version of the datasource to promote maximum backwards compatibility.


srclosson:
Just looking for a second opinion before merging that's all. Anyone? It should be quick!

PR #45, opened 2019-11-07
fixed issues in readme.md

cuchalin:
Removed unnecessary 'is' in the sentence, 'This is should return a comma separated string of tags'
Fixed mistakes in following sentence.
               Instead of hard-coding things like server, application and sensor name in your metric queries 
               you can use variables in their place.
To this
              Instead of hard-coding things like servers, application and sensor name in your metric queries 
              can use variables in their place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,7 +260,7 @@ To make writing queries easier there are two Grafana macros that can be used in 
 
 ## Templating with Variables
 
-Instead of hard-coding things like server, application and sensor name in your metric queries can use variables in their place. Variables are shown as dropdown select boxes at the top of the dashboard. These dropdowns make it easy to change the data being displayed in your dashboard.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -260,7 +260,7 @@ To make writing queries easier there are two Grafana macros that can be used in
 
 ## Templating with Variables
 
-Instead of hard-coding things like server, application and sensor name in you metric queries you can use variables in their place. Variables are shown as dropdown select boxes at the top of the dashboard. These dropdowns make it easy to change the data being displayed in your dashboard.
+Instead of hard-coding things like server, application and sensor name in your metric queries can use variables in their place. Variables are shown as dropdown select boxes at the top of the dashboard. These dropdowns make it easy to change the data being displayed in your dashboard.
```
rxlecky: The second "you" should not be deleted, i.e. "... and sensor name in your metric queries YOU can use variables..." is correct.

PR #49, opened 2019-11-27
Make build targets more explicit for go build

sputnik13:
Added GOOS and GOARCH to each go build statement in Makefile to make
target OS and ARCH more explicit and allow cross-compilation.


briangann:
we've switched over to using mage now, the makefile method is going away.

PR #67, opened 2020-02-15
remove *.d.ts hack

ryantxu:
I updated toolkit base config so it does not produce the extra files... now the hack should not be necessary

https://github.com/grafana/grafana/commit/21b53b7d79034ea7e5859feb539c7565c22b2c6a


ryantxu:
Close in favor of #78

Issue #61, opened 2020-01-20
Variables of type "Query" are broken on v2.0.1 (for ADX data source)


gilkgit:
Variables of type "Query" are broken on v2.0.1 (for ADX data source)

User error:
"
Templating
Template variables could not be initialized: undefined
"

chrispellett:
I'm having this same problem with both v2.0.3 and v2.0.4 (of the ADX data source). We have our plugin run in a Hosted Grafana instance that is brand new (as of a few days ago)

I am unable to create a dashboard variable with the "Query" type. We can create dashboard visualizations, so we know our problem isn't an "ADX connectivity" issue (or related).

Whenever I write any sort of query, I am met with the popup notification:
"Templating
Template variables could not be initialized: undefined"
(just like OP)

When I try and dig in to the error (on the client side) by looking at dev tools, I only see a generic "There was an error" reported in the console.

The error appears to be from the /api/tsdb/query endpoint returning a 415 status and the response body:
[{"classification":"ContentTypeError","message":"Empty Content-Type"}]

Unfortunately, I'm not sure where this error is originating from to dig in any more.

Any ideas on what to look for?

gilkgit:
I wish we could vote for priorities. For me, that's the only blocking bug that currently exists.

sunker:
@gilkgit can you still reproduce this? I think this was fixed in #79

gilkgit:
It's been fixed a while ago. You're welcome to close it. Thanks!

sunker:
Thanks @gilkgit!

PR #79, opened 2020-03-11
Fix for variable query to backend

srclosson:
Why: This is an interin fix for variable queries to work. Done for Microsoft in response to #61


src/datasource.ts
```diff
@@ -254,10 +263,10 @@ export class KustoDBDatasource {
   }
 
   private buildQuery(query: string, options: any, database: string) {
-    if (typeof options === 'undefined') {
+    if (!options) {
```
scottlepp: options = options ?? {}

PR #95, opened 2020-07-13
WIP: rework query editor 

dprokop:


daniellee:
# TODO

- [ ]  Readonly Query text area
- [ ] Fix the typing errors for DataSourceWithBackend
- [ ] Operators for the where filter
- [ ] Add relevant reduce functions
- [ ] Handle non-aggregated queries (partially done)
- [ ] Switch between visual and raw queries

jessover9000:
Hello there, I'm testing the status quo UX now. it's already looking and working quite well. I have a bunch of feedback:

- Database/From dropdowns: It seems a bit weird to me that I can't use backspace to delete the currently selected "From". It's not a common pattern for such a combo box that I have type something with the current value still fully displayed in the input, and then it starts showing what I typed from an empty field. not sure whether that's a general component issue though. I guess that might be what @daniellee meant when he wrote on Slack that "The Select component is not very easy to write custom values in"
- where dropdown: When I type something in there, e.g. "de", it will show stuff starting with "C" that contains "de" before showing stuff that actually starts with "de". I guess we can do it that way, but in that case we should highlight the "de" in the search result because otherwise the user, who might expect words starting with D to show up first, might be confused.
- operator dropdown: I'd also suggest just choosing a default value and using that, any placeholder text would take a lot of space and cause the UI to jump, but that empty button looks just plain weird. I also think that the dropdown text that appears when you click the operator button before choosing a column shouldn't say "No options available" but "Please choose a column first." And you already know this, but the dropdown behaviour for the operator button right now is hard to navigate, can we make it wider and not be cut off by the drawer? 🤔 
Screenshot 2020-07-29 at 13 41 09
Screenshot 2020-07-29 at 13 43 53
- Buttons: there should only be one primary action button, not two. I'm leaning towards the "Run query" one. The run query button also needs better copy, maybe we can change the text to "Test query"? I think the "edit KQL" button and a potential copypaste button for the query code should be grouped together away from the "Run query" button. Maybe the "Edit KQL" and a copy button could be moved to the top right corner of the query dropdown but under the bar with the moving and deleting icons, and the "Run query" can stay pretty much where it is.
- Flipping between modes: I think the button solution is fine. What's a bit problematic here is that on clicking "Edit KQL" the view jumps and changes to something completely different. As a user, I'm a bit overwhelmed trying to make the transfer between my visual query and the code editor, for all I know it could be something completely different in there – I can't visually process the change very well and there is no relation between the modes. We could do an animation of some sort where the visual editor just fades away, the "Format as" bar moves up and the code becomes editable. Also we could arrange the fields and buttons to be more similar to the visual editor, meaning the "Run" and "Edit with Query Builder" buttons are in the same places for both modes and maybe even the "Format as" and "Alias by" moves to above the actual code editor. If that makes sense to you I can draft a design for it. 
- Code edit mode itself: Using the code editor, I kept wanting to click the "Query builder" button below the code to run my queries. I feel like the top position for the "Run" button isn't intuitive when your natural way of using and reading the code is downwards. It doesn't help that they are both formatted as primary buttons. I'd suggest moving the run button down and putting the query builder button in the top right corner for both modes.
- "Losing" your visual query in code edit mode: I'm not sure where the point is where you actually lose your query. When I'm not changing the code but just flipping between code and visual editor, of course I don't lose anything. Does it happen as soon as you enter anything there? If yes, I would definitely show some visual feedback about this. Question to you devs though: Do you think it should be a modal where the user has to click "Yes I'm sure" or "No, take me back to the visual editor" or is that too much of a disruption and we should try something that's noticeable but doesn't take the user out of what they're doing right now?
- Error messages in the visual query editor: If you click on "Run query" and nothing happens and you only get "No data", I think an error message is better than no feedback at all.

mckn:
Closing this one in favor of #101

PR #109, opened 2020-08-12
wip: timeout on queries

kylebrandt:


kylebrandt:
Actually not needed, issue was needing to increase timeout configuration for reporting.

PR #130, opened 2020-09-14
Small fixes

hugohaggmark:
This PR addresses the following:
- Where statement - ~in should have a multiselect as input type.
- Introduce a "max number of returned values" setting for the datasource. Currently it is fixed to 10 000 which seems to be too low in some scenarios.

Before moving on, I'd like feedback on moving the KustoExpressionParser to the datasource :)

### DIFF
--- a/src/partials/config.html
+++ b/src/partials/config.html
@@ -108,7 +108,7 @@
       <span class="gf-form-label width-11">Default take limit</span>
       <input type="number" min="1" class="gf-form-input width-9" ng-model="ctrl.current.jsonData.defaultTakeLimit" spellcheck="false" placeholder="10000"></input>
       <info-popover mode="right-absolute">
-        The default row limit is 10 000. If you need a greater or lower default, enter the required value.
+        The default take limit is 10 000. If you need a greater or lower default, enter the required value.
       </info-popover>
     </div>
   </div>
--- a/src/partials/config.html
+++ b/src/partials/config.html
@@ -105,7 +105,7 @@
   </div>
   <div class="gf-form-inline">
     <div class="gf-form">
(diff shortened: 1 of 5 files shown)


src/types.ts
```diff
@@ -49,6 +49,7 @@ export const defaultQuery: Pick<KustoQuery, 'query' | 'expression' | 'querySourc
 export interface AdxDataSourceOptions extends DataSourceJsonData {
   defaultDatabase: string;
   minimalCache: number;
+  defaultTakeLimit: number;
```
hugohaggmark, line 52: Not sure about the naming here :)

src/partials/config.html
```diff
@@ -103,6 +103,15 @@ <h5>3. Configure the connection in Grafana</h5>
       </info-popover>
     </div>
   </div>
+  <div class="gf-form-inline">
+    <div class="gf-form">
+      <span class="gf-form-label width-11">Default row limit</span>
```
hugohaggmark, line 108:
```suggestion
      <span class="gf-form-label width-11">Default take limit</span>
```

src/partials/config.html
```diff
@@ -103,6 +103,15 @@ <h5>3. Configure the connection in Grafana</h5>
...
+      <input type="number" min="1" class="gf-form-input width-9" ng-model="ctrl.current.jsonData.defaultTakeLimit" spellcheck="false" placeholder="10000"></input>
+      <info-popover mode="right-absolute">
+        The default row limit is 10 000. If you need a greater or lower default, enter the required value.
```
hugohaggmark, line 111:
```suggestion
        The default take limit is 10 000. If you need a greater or lower default, enter the required value.
```

PR #132, opened 2020-09-15
DatasourceSettings: Adds default editor mode

hugohaggmark:
Adds a new setting to the data source config page:
image

Introduced `EditorMode` enum and refactored some magic strings so please check they're ok :)


mckn (approved):
🙌

PR #133, opened 2020-09-15
alerting: Add fromAlerting to identify queries form alerting

kylebrandt:
depends on grafana/grafana#27599 (Edit: Now merged into master, should be in next 7.2 beta)

will be used so the macros will not be expanded in "optimized query" style for alerts

cc @mckn 

Creating as draft since it isn't used yet + the dependency is not merged yet. Once merged, for pre 7.2 versions of Grafana, this will always be false.


mckn (approved):
LGTM!

andresmgot:
Closing this due to inactivity.

PR #134, opened 2020-09-16
Chore: Adds eslint

hugohaggmark:
This PR adds eslint config and extends the Grafana eslint config. There were a 18 warnings and 1 error all related to React Hooks that I've fixed. I'll leave comments in the files where I'm not sure why we're missing dependencies or other issues.

### DIFF
--- a/src/QueryEditor.tsx
+++ b/src/QueryEditor.tsx
@@ -16,7 +16,7 @@ import { needsToBeMigrated, migrateQuery } from 'migrations/query';
 type Props = QueryEditorProps<AdxDataSource, KustoQuery, AdxDataSourceOptions>;
 
 export const QueryEditor: React.FC<Props> = props => {
-  const { datasource } = props;
+  const { datasource, onChange, onRunQuery, query } = props;
   const executedQuery = useExecutedQuery(props.data);
   const executedQueryError = useExecutedQueryError(props.data);
   const dirty = useDirty(props.query.query, executedQuery);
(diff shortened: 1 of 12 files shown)


src/QueryEditor.tsx
```diff
@@ -41,24 +42,25 @@ export const QueryEditor: React.FC<Props> = props => {
       props.onRunQuery();
     }
   }, []);
```
hugohaggmark, line 44: I guessed that we only want to check this once on component mount?

src/editor/components/field/QueryEditorFieldSection.tsx
```diff
@@ -40,11 +32,18 @@ export const QueryEditorFieldSection = (config: FieldSectionConfiguration): Reac
...
     );
   };
 };
+
+const onChange = (propsOnChange: (value: QueryEditorExpression) => void) => (property: QueryEditorProperty) => {
+  propsOnChange({
+    type: QueryEditorExpressionType.Property,
+    property,
+  } as QueryEditorPropertyExpression);
+};
```
hugohaggmark, line 49: The only way I figured that we could solve the eslint error with was to extract a proper function

src/editor/components/groupBy/QueryEditorGroupBy.tsx
```diff
@@ -47,9 +49,10 @@ export const QueryEditorGroupBy: React.FC<Props> = props => {
         interval,
       };
 
-      props.onChange(payload);
+      onChange(payload); // adding onChange to dependency array below causes maximum call depth error
```
hugohaggmark, line 52: I wrote a comment about this one. If I add onChange to dependencies we get a recursion with a maximum call stack depth. Didn't look into this further, but we really should dive deeper into this when we have time.

src/editor/components/reduce/QueryEditorReduce.tsx
```diff
@@ -74,9 +76,10 @@ export const QueryEditorReduce: React.FC<Props> = props => {
         parameters: parameters,
       };
 
-      props.onChange(payload);
+      onChange(payload); // adding onChange to dependency array below causes maximum call depth error
```
hugohaggmark, line 79: I wrote a comment about this one. If I add onChange to dependencies we get a recursion with a maximum call stack depth. Didn't look into this further, but we really should dive deeper into this when we have time.

src/QueryEditor.tsx
mckn: yes 👍

src/editor/components/field/QueryEditorFieldSection.tsx
mckn: Feels like a good solution! The other one is only an optimization that can be skipped for now.

src/editor/components/groupBy/QueryEditorGroupBy.tsx
mckn: hmm, it looks like we are doing something really bad here and it should probably be changed. Lets add this one to the list. I'm not really following what we are trying to solve so need to have a look at that before I can suggest a solution.

mckn (approved):
It looks really good!

We need to have a look at how we manage state with hooks in the two examples that you added comments for below.

Issue #125, opened 2020-09-11
Query performance improvements by enabling query caching


mckn:
To be able to increase the performance of making queries against ADX data sources with a massive amount of data we would like to add the possibility to enable query caching via the datasource settings.

The following things should be done:

- [x] When translating the time macros we should start using datetime operator in KQL. Also we should round the time values in the query to increase the chance of hitting the cache. Timestamp rounding should be configurable via the datasource settings. See example on implementation here: https://github.com/erik-jansen/azure-data-explorer-datasource/commit/b5817d64e24d37baaadbfadc73873931aaafc352 (#140)
- [x] We should make is possible to set the query cache max age for the queries via the datasource settings. See example on implementation here: https://github.com/erik-jansen/azure-data-explorer-datasource/commit/351c6f651957c1f440409307a71e1c5ef176b904  (#131)
- [x] There should also be a setting for toggling weak consistency when doing queries. See example on implementation here: https://github.com/erik-jansen/azure-data-explorer-datasource/commit/351c6f651957c1f440409307a71e1c5ef176b904  (#131)

More information about caching in ADX can be found here: https://docs.microsoft.com/en-us/azure/data-explorer/kusto/query/query-results-cache

PR #140, opened 2020-09-16
Dynamic caching logic

mckn:
This PR adds dynamic caching functionality that will apply caching logic based in the query and time range.

Closes: #125

Skärmavbild 2020-09-16 kl  22 41 06


hugohaggmark (approved):
Frontend looks good 🌟

mckn:
I am waiting on a couple of answers from the people at MS regarding some of the edge cases. Will add those as tests when I get a reply.

pkg/azuredx/cache_test.go
```diff
@@ -0,0 +1,180 @@
...
+				},
+			},
+		},
+		{
+			name: "should adjust settings based on query bin size 1h",
+			configuration: dataSourceData{
+				CacheMaxAge:    "10s",
+				DynamicCaching: true,
+			},
+			query: `SampleData\n` +
```
kylebrandt, line 71: I think the diff ends up funny when this fails with using `+`, so better to use three backticks instead

pkg/azuredx/cache.go
```diff
@@ -0,0 +1,111 @@
...
+	if diff > time.Hour {
+		return time.Minute
+	}
+
+	return cacheMinAge
+}
+
+// binSizeRE finds the bin size of a query by looking at the summarize statement
+// e.g. summarize avg(Column), count(Column) by bin(TimeColumn, 1d)
+var binSizeRE = regexp.MustCompile(`summarize [\w\$\(\)\.,\s]+ by bin\([\w\$\(\)\.]+, ` +
```
kylebrandt, line 93: Worry about change query behavior based on a regex. But if we have to do it, I think we should just detect the bin.

kminehart:
out of curiosity, we have a couple of enterprise datasources that have caching built in; usually it's just an in-memory map. Is this a similar thing or some cache options provided by Azure?

mckn:
> out of curiosity, we have a couple of enterprise datasources that have caching built in; usually it's just an in-memory map. Is this a similar thing or some cache options provided by Azure?

This is basically just setting cache headers that are being sent to Azure that will trigger them to cache queries. But what we have added is the possibility to either set these values manually or enabling our datasource plugin to detect what caching options would be optimal on a per-query basis.

PR #142, opened 2020-09-18
Added first simple version of timeshift

mckn:


src/components/VisualQueryEditor.tsx
```diff
@@ -69,12 +70,20 @@ export const VisualQueryEditor: React.FC<Props> = props => {
...
+          expression: query.expression,
+          index,
+        },
+        tableSchema.value
+      );
+
       return values.map(value => ({ value, label: value }));
     },
-    [datasource, databaseName, tableName, tableSchema.value]
+    [datasource, databaseName, tableSchema.value, query.expression]
```
hugohaggmark, line 86: nit: You should extract autoCompleteQuery from datasource and add it as a dependency.
```suggestion
    [autoCompleteQuery, databaseName, tableSchema.value, query.expression]
```

src/components/VisualQueryEditor.tsx
```diff
@@ -175,10 +184,33 @@ export const VisualQueryEditor: React.FC<Props> = props => {
...
+      };
+
+      onChangeQuery({
+        ...query,
+        resultFormat: resultFormat,
+        database: database,
+        expression: next,
+      });
+    },
+    [onChangeQuery, query, resultFormat, database, table]
```
hugohaggmark, line 207: nit: Missing defaultQuery.expression and query.expression in dependencies.
```suggestion
    [onChangeQuery, query, resultFormat, database, table]
```

hugohaggmark (approved):
Great work on the tests and the refactor, I just left some small nits that you might want to consider.

src/components/VisualQueryEditor.tsx
mckn: hmm, the defaultQuery is immutable and should not be changed.

src/components/VisualQueryEditor.tsx
mckn: Great catch!

src/components/VisualQueryEditor.tsx
mckn: Looks like react-hooks eslint recommend not to add them since changing them will not result in a re-render.

PR #143, opened 2020-09-21
Refactor: switches to use AsyncSelect instead of Select

hugohaggmark:
This change does not affect Selects that already have auto completion. My thinking is those will have to be changed in another PR.


mckn (approved):
I think this is amazing! Great work!

PR #145, opened 2020-09-21
Autocomplete should populate some of the values prior to typing

mckn:


src/components/VisualQueryEditor.tsx
```diff
@@ -71,7 +71,7 @@ export const VisualQueryEditor: React.FC<Props> = props => {
 
   const onAutoComplete = useCallback(
     async (index: string, search: QueryEditorOperatorExpression) => {
-      const values = await autoCompleteQuery(
+      const values = await datasource.autoCompleteQuery(
```
hugohaggmark, line 74: nit: in this case it doesn't matter but it's considered best practices to destruct properties on objects used as dependencies in hooks

hugohaggmark (approved):
LGTM and works as expected!

PR #147, opened 2020-09-22
Refactor: Fixes so backspace removes selected value

hugohaggmark:


mckn:
I think this works great! It would be even better if you could make it select the "previously selected" if you clear but do not select anything new (as in the gif). Maybe timebox and see if it is possible to get in place?

mckn:
Kapture 2020-09-22 at 07 41 48

hugohaggmark:
Will have a quick look, will probably introduce ugly hack :)

hugohaggmark:
Just a heads up the gif animation is for selects that I haven't changed.
https://github.com/grafana/azure-data-explorer-datasource/pull/147#issuecomment-696519240

hugohaggmark:
@mckn please have another look, this is just woking for QueryEditorField right now

mckn:
> @mckn please have another look, this is just woking for QueryEditorField right now

This is spot on! Awesome, we should probably add this for the operator value field as well.

mckn (approved):
LGTM!

PR #152, opened 2020-09-22
Excluding empty operators/where values from generated query

### LINKED ISSUE #128: When added a where filter but not yet triggered a query the following error occur.
image004

mckn:
Fixed #128


hugohaggmark (approved):
LGTM!

PR #155, opened 2020-09-23
Bugfix: fixes during testing session

mckn:
- We fixed so you have a simple input for dateTime and number values.
- Made sure we have proper mappings between column types and operators.
- No auto complete will trigger for dateTime or number values.


hugohaggmark (approved):
LGTM!

PR #170, opened 2020-10-12
feat(conversions): add conversion for decimal

### LINKED ISSUE #116: Decimal column support
Using Grafana 7.1.4 with ADX plugin 3.0.1.. If I run any query against ADX that returns a decimal, I get this error:

image

secustor:
Adds a conversion for decimal values. 

Closes #116


secustor:
@mckn Can you point to some documentation to fix this error? 
I have build the plugin using `grafana-toolkit plugin:build` and then copied the content in a folder named `azure-data-explorer-datasource` under `/var/lib/grafana/plugins/`.

image

Thx!

secustor:
This works for me now. 
image

kylebrandt (approved):
LGTM.

One concern would be properly handling the 128 bit width.

That would involved:
 - Actually understanding https://golang.org/pkg/math/big/#Float , which I honestly do not.
 - Maybe understand this part of the go arrow sdk https://github.com/apache/arrow/blob/487895fe10540488f99d7d26f0a3b5e77c097122/go/arrow/decimal128/decimal128.go, so it can go into grafana's go sdk
 - Look into what arrow JS lib would do with these numbers, which would probably turn them into a float anyways

So all that seems a bit futile when this will just make things work for most people, and we can look into large number precision if anyone ever files an issue and actually has a problem.

PR #175, opened 2020-10-21
Fix: this PR is fixing broken build by cleaning up the decimal code a bit.

mckn:
#170

### DIFF
--- a/pkg/azuredx/table.go
+++ b/pkg/azuredx/table.go
@@ -249,9 +249,19 @@ var decimalConverter = data.FieldConverter{
 		if v == nil {
 			return af, nil
 		}
+
+		jS, ok := v.(string)
+		if ok {
+			out, err := strconv.ParseFloat(jS, 64)
+			if err != nil {
+				return nil, err
+			}
+			return out, nil
+		}
+
 		jN, ok := v.(json.Number)
 		if !ok {
-			return nil, fmt.Errorf("unexpected type, expected json.Number but got type %T with a value of %v", v, v)
+			return nil, fmt.Errorf("unexpected type, expected json.Number or string but got type %T with a value of %v", v, v)
 		}
 		out, err := jN.Float64()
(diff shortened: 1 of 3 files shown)


pkg/azuredx/table.go
```diff
@@ -245,15 +245,25 @@ var longConverter = data.FieldConverter{
...
-		if !ok {
-			return nil, fmt.Errorf("unexpected type, expected String but got type %T with a value of %v", v, v)
+
+		jS, sOk := v.(string)
+		if sOk {
+			out, err := strconv.ParseFloat(jS, 64)
+			if err != nil {
+				return nil, err
+			}
+			return &out, err
```
kylebrandt, line 259:
```suggestion
			return &out, nil
```

pkg/azuredx/table.go
```diff
@@ -245,19 +245,29 @@ var longConverter = data.FieldConverter{
...
+		var af *float64
 		if v == nil {
-			return ai, nil
+			return af, nil
 		}
-		jN, ok := v.(string)
-		if !ok {
-			return nil, fmt.Errorf("unexpected type, expected String but got type %T with a value of %v", v, v)
+
+		jS, sOk := v.(string)
```
kylebrandt, line 253: Is there a case where it would be a string?

in `tableFromJSON` we use https://golang.org/pkg/encoding/json/#Decoder.UseNumber, so it should also be a json.Number I believe.

mckn: When running it live we get it as a string. Not sure why but it is 🙈

mckn: Try this query: `datatable (Value:decimal)[1.2,22.1111,2323323.3] | where Value > 2`

kylebrandt: Perhaps the test data is wrong and it is always a string? That would make sense for this type...

Issue #168, opened 2020-10-08
Query Builder: generates KQL invalid query ("by ,Type")


sloutsky:
When trying to create new Dashboard template - incorrect KQL query is generated (extra comma):

Failed to execute query: Syntax error: Query could not be parsed: SYN0002: A recognition error occurred. [line:position=3:23]. Query: 'EventsFromLiveStream
| where CreatedAt >= datetime(2020-10-08T03:39:45Z) and CreatedAt <= datetime(2020-10-08T09:40:00Z)
| summarize count() by , Type
| take 10000'

![image](https://user-images.githubusercontent.com/17802887/95441822-46d97200-0963-11eb-97c1-6482653a25ac.png)

Issue #160, opened 2020-09-28
Value column is included in the "by" clause


avneraa:
When using the "count()" aggregation the column appears in the "by" clause, this is wrong:

![image](https://user-images.githubusercontent.com/4261958/94426112-bf874400-0195-11eb-9a08-60ddac08603f.png)

mckn:
@avneraa great feedback, how should the query be formatted in this scenario?

avneraa:
The value columns should not be included in the "by" clause, you need to simply remove the code that adds it to it. As a side not, I find the name "value columns" really misleading, I think that calling this "summarize" would be much clearer and look much more natural, especially if you remove the "(summarize)" from the "group by" area, so it reads like the query: |summarize ... by ...

avneraa:
Hi Marcus, I updated the two issues with the answers.

Issue #159, opened 2020-09-28
When the user just want to count, the query builder gives an error


avneraa:
When user wants to just show the count, they do not need to provide any specific column to count, this gives an error:
![image](https://user-images.githubusercontent.com/4261958/94425940-733c0400-0195-11eb-8799-09b122d9f0cb.png)

mckn:
hmm, I see. You should not need to specify a column to count. I will add this as a valid option but I think in the long run we should probably improve the UI so this feels more natural in the UI.

avneraa:
I would consider flipping the order, so the drop down shows all the relevant aggregation functions and the next blue label says "Column" and the dropdown shows the list of relevant columns. So if the user selects "count" then the column list is empty.  One huge benefit would be that it will be closer to how the user would write the query "summarize dcount(Col1)" so looking at the generated query is much more intuitive. 
As a side note, I suggested to rename "value columns" to "summarize" and remove the "summarize" from the "group by" label.

PR #178, opened 2020-10-22
UI: improved the query builder so make aggregations less confusing

mckn:
Fixes: #168, #160, #159

aggregate_by


mckn:
@sloutsky @avneraa @Gabilehner can you verify that this looks as expected.

avneraa:
Looks good. I would remove the "by" so the label should just be "Aggregate". Other than this looks much clearer!

Issue #183, opened 2020-10-27
HTTP client should enforce a timeout


marefr:
The HTTP client in use seems to use the default of what Go provides. This is not optimal for production usage, see for example https://medium.com/@nate510/don-t-use-go-s-default-http-client-4804cb19f779, "Go’s http package doesn’t specify request timeouts by default, allowing services to hijack your goroutines.". 

Here's the code in question that gets a HTTP client. Not sure yet if it's uses Go's default HTTP client since it gets it from the oauth2 library:
https://github.com/grafana/azure-data-explorer-datasource/blob/015898bee2f69b557ec7dd85c648888c47104504/pkg/azuredx/azuredx.go#L119

One outstanding question is if this should be configured per plugin or per plugin data source instance on the data source configuration page. Maybe both, default timeout on global level with possibility to override on a data source instance by data source instance basis. This is maybe related to #165 that suggests adding a timeout to the data source configuration page and also discusses how to let service know how long to execute a request before it timeout.

kylebrandt:
Would probably be good to set server side timeout to be a little less than our timeout as well, so get clean error messages and don't leave queries running server side.

Abandoned Draft PR of this before connection properties was added, should be able to add it to current properties I think:
 - https://github.com/grafana/azure-data-explorer-datasource/pull/109/files

Draft PR for similar issue for server side timeout in https://github.com/grafana/grafana/pull/28211

PR #184, opened 2020-10-28
Query: adding datasource settings for query timeout.

mckn:
For https://github.com/grafana/azure-data-explorer-datasource/issues/183

- Add data source setting queryTimeout, default is 30 seconds
- Set API client timeout to queryTimeout+5 seconds
- Set server execution timeout via connection property object to queryTimeout

Based on my (Kyle) testing:
 - If you make a CPU expensive query, you get a "Watchdog" error from Azure
 - If you make a quick query that returns a lot of data, you may hit our client timeout while receiving the data
 - Query that do not hit the timeout continue to work normally.

I think the client for auth may still have no timeout.

### DIFF
--- a/pkg/azuredx/azuredx.go
+++ b/pkg/azuredx/azuredx.go
@@ -38,17 +38,25 @@ func (qm *QueryModel) Interpolate() (err error) {
 // dataSourceData holds the datasource configuration information for Azure Data Explorer's API
 // that is needed to execute a request against Azure's Data Explorer API.
 type dataSourceData struct {
-	ClientID           string        `json:"clientId"`
-	TenantID           string        `json:"tenantId"`
-	ClusterURL         string        `json:"clusterUrl"`
-	DefaultDatabase    string        `json:"defaultDatabase"`
-	Secret             string        `json:"-"`
-	DataConsistency    string        `json:"dataConsistency"`
-	CacheMaxAge        string        `json:"cacheMaxAge"`
-	DynamicCaching     bool          `json:"dynamicCaching"`
-	QueryTimeoutRaw    string        `json:"queryTimeout"`
-	QueryTimeout       time.Duration `json:"-"`
-	ServerTimeoutValue string        `json:"-"`
+	ClientID        string `json:"clientId"`
+	TenantID        string `json:"tenantId"`
+	ClusterURL      string `json:"clusterUrl"`
+	DefaultDatabase string `json:"defaultDatabase"`
+	Secret          string `json:"-"`
+	DataConsistency string `json:"dataConsistency"`
+	CacheMaxAge     string `json:"cacheMaxAge"`
+	DynamicCaching  bool   `json:"dynamicCaching"`
+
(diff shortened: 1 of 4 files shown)


src/partials/config.html
```diff
@@ -98,6 +98,15 @@ <h3 class="page-heading">Connection details</h3>
 
 <h3 class="page-heading">Query Optimizations</h3>
 <div class ="gf-form-group">
+  <div class="gf-form-inline">
+    <div class="gf-form">
+      <span class="gf-form-label width-11">Query timeout</span>
```
kylebrandt, line 103: @mckn Min 1s, Max is 1h, can you adjust the UI to match that?

mckn: Yupp - I fix :)

kylebrandt:
We may consider using instance manager in another iteration to reuse clients per datasource instance within the plugin https://pkg.go.dev/github.com/grafana/grafana-plugin-sdk-go@v0.78.0/backend/instancemgmt

kylebrandt:
Note: I am able to pretty much break my local instance on master, or this with a high timeout. I am pretty sure, by looking at `.show running queries` that I'm hitting the concurrency limits in azure https://docs.microsoft.com/en-us/azure/data-explorer/kusto/concepts/querylimits#limit-on-query-concurrency

But by having this and a shorter timeout. This should allow users better control of concurrency because they can limit the lifetime of queries in the system with this setting.

pkg/azuredx/azuredx.go
```diff
@@ -116,10 +146,49 @@ func NewClient(ctx context.Context, dInfo *backend.DataSourceInstanceSettings) (
...
+		Transport: authClient.Transport,
+		// We add five seconds to the timeout so the client does not timeout before the server.
+		// This is because the QueryTimeout property is used to set the server execution timeout
+		// for queries. The server execution timeout does not apply to retrieving data, so when
+		// a query returns a large amount of data, timeouts will still occur while the data is
+		// being downloaded.
+		// In the future, if we get the timeout value from Grafana's data source proxy setting, we
+		// may have to flip this to subtract time.
+		Timeout: c.dataSourceData.QueryTimeout + 5*time.Second,
+	}
```
marefr, line 167: Looked at the library code and this should work. However, is there any difference to do similar to what you already have? I think `conf.TokenSource(ctx)` refreshes the token if needed so doing this should be the same thing.

```go
c.Client = conf.Client(ctx)
s.Client.Timeout = c.dataSourceData.QueryTimeout + 5*time.Second
```

But in my head shouldn't make a difference only that you would create two HTTP clients instead of one.

marefr (approved):
Added a comment, but shouldn't be a blocker. 

LGTM

pkg/azuredx/azuredx.go
kylebrandt: I did this because the comment on Config.Client says: `// The returned Client and its Transport should not be modified.`, where NewClient I think is for that maybe?

marefr: Make sense :+1:

Issue #166, opened 2020-10-02
Feature Request: Pass Grafana userid to ADX as a client request property


brycefisher:
**Use Case**
 - I want to audit which user is sending queries (say from the "Explore" page in Grafana) to notify users about inefficent/dangerous queries
 - Grafana sends some representation of the user executing the query with the request as a client request property
 - Monitoring downstream can identify the user and trigger a slack message or other notification to that user if high impact queries are detected

**Notes**
 - Key spot in Adx package: https://github.com/grafana/azure-data-explorer-datasource/blob/f26d6f3d4fdc4a1b9c09c1305a2ccb5b3a0d8f92/pkg/azuredx/azuredx.go#L148
 - Documentation on [`User` named property](https://docs.microsoft.com/en-us/azure/data-explorer/kusto/api/netfx/request-properties#the-user-x-ms-user-named-property)

**Challenges**
 - I'm uncertain if Grafana provides this information to all plugins by default, or if this would require changes to the main server application
 - Can Grafana easily represent alert queries' user somehow already?

gilkgit:
In addition to this ticket (Which I think is a duplicate of an older open ticket) - 
It will also be useful to add the the dashboard name to the Kusto request, so we'll be able to easily find the offending dashboard from Kusto

brycefisher:
Overall, I think that makes sense. I would favor dashboard ID or url, since users can rename the panel and I don't believe there are uniqueness guarrantees on dashboard names.

Also, we could pass through the panel ID with the dashboard ID.

For all the functionality above, I would look at altering how request ID's activity name is generated in above linked method.

mckn:
Great idea and something we easily could add!

brycefisher:
@mckn if I wanted to contribute a PR for this, are there any pointers you could provide me?

brycefisher:
@mckn would you accept a well written PR for this feature? It seems like something I could possibly figure out and would be extremely valuable to me personally :-)

joshhunt:
@brycefisher Sure, go ahead! Let us know if you have any questions.

brycefisher:
So, I starting playing with this and ran into a couple issues:

[...]

PR #196, opened 2021-01-10
Enhancement: Add more visibility into Grafana using x-ms-client-request-id

brycefisher:
Initial attempt at low hanging fruit from #166

### DIFF
--- a/pkg/datasource.go
+++ b/pkg/datasource.go
@@ -13,7 +13,7 @@ import (
 // GrafanaAzureDXDatasource stores reference to plugin and logger
 type GrafanaAzureDXDatasource struct{}
 
-func (plugin *GrafanaAzureDXDatasource) handleQuery(client *azuredx.Client, q backend.DataQuery) backend.DataResponse {
+func (plugin *GrafanaAzureDXDatasource) handleQuery(client *azuredx.Client, q backend.DataQuery, user string) backend.DataResponse {
 	resp := backend.DataResponse{}
 	qm := &azuredx.QueryModel{}
 
@@ -56,7 +56,7 @@ func (plugin *GrafanaAzureDXDatasource) handleQuery(client *azuredx.Client, q ba
 		CSL:        qm.Query,
 		DB:         qm.Database,
 		Properties: azuredx.NewConnectionProperties(client, cs),
-	}, qm.QuerySource)
+	}, qm.QuerySource, user)
 	if err != nil {
 		backend.Logger.Debug("error building kusto request", "error", err.Error())
 		errorWithFrame(err)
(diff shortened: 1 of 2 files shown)


pkg/datasource.go
```diff
@@ -143,8 +143,13 @@ func (plugin *GrafanaAzureDXDatasource) QueryData(ctx context.Context, req *back
 		return nil, err
 	}
 
+	email := "Noninteractive"
+	if req.PluginContext.User != nil {
+		email = req.PluginContext.User.Email
```
brycefisher, line 148: @joshhunt @mckn It looked to me like this should work, but inside the docker-compose container, with only the admin user, this always resulted in either `nil` or `"" (I'm extremely rusty with go at this point and can't remember the semantics). Is there a more realistic way to test this?

sunker: When investigating this I managed to find a bug in Grafana. The PluginContext.User.Email is not set correctly. I've created a [pr](https://github.com/grafana/grafana/pull/30249) for that, but unfortunately that won't be available to the adx plugin until next release of Grafana. Can this PR wait for that? If not, we can use PluginContext.User.Login or PluginContext.User.Name. At least for [Grafana auth](https://grafana.com/docs/grafana/latest/auth/overview/#grafana-auth), PluginContext.User.Login will always be set. 

Thanks for finding the bug!

sunker:
For privacy and security reasons, I don't think it's quite right to pass information about the user in a header like that by default. I think it should be up to the user to enable this feature. I'll investigate how to makes this configurable in a plugin and report back here.

sunker:
> For privacy and security reasons, I don't think it's quite right to pass information about the user in a header like that by default. I think it should be up to the user to enable this feature. I'll investigate how to makes this configurable in a plugin and report back here.

I'll do this in a separate PR as soon as this is merged!

pkg/datasource.go
sunker: Actually, I think we should always use PluginContext.User.Login since it will always be set. So please update the code to use that instead.

brycefisher: Glad I could help with the bug discovery!

Sounds like a plan! I think any distinct value per user would be a major improvement. I'll push up this change momentarily...

brycefisher: @sunker how does this look now?

pkg/datasource.go
```diff
@@ -142,8 +142,13 @@ func (plugin *GrafanaAzureDXDatasource) QueryData(ctx context.Context, req *back
 		return nil, err
 	}
 
+	email := "Noninteractive"
```
sunker, line 145: should be userLogin now

sunker (changes requested):
Looks good! Just one minor thing. Fix that and then I'll merge.

pkg/datasource.go
sunker: Actually I can fix that in the opt in pr. I'll merge. Again thanks for contributing!

sunker (approved):

pkg/datasource.go
brycefisher: Thank you !!! Yes, I totally missed renaming the variable -- would have been very confusing.

brycefisher:
@joshhunt @mckn @sunker how soon do you think a new release could be cut with this PR included?

PR #198, opened 2021-01-13
Chore: Use jsoniter instead of encoding/json

aknuds1:
Use [jsoniter](https://github.com/json-iterator/go) instead of encoding/json for better JSON decoding/encoding perf.


papagian (approved):
LGTM

PR #199, opened 2021-01-14
Make username tracking optional

sunker:
User name tracking will be disabled by default, but can be activated in the plugin config page. 

Screenshot from 2021-01-14 13-28-46

### DIFF
--- a/pkg/datasource.go
+++ b/pkg/datasource.go
@@ -142,13 +142,13 @@ func (plugin *GrafanaAzureDXDatasource) QueryData(ctx context.Context, req *back
 		return nil, err
 	}
 
-	email := "Noninteractive"
+	username := "Noninteractive"
 	if req.PluginContext.User != nil {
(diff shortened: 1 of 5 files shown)


pkg/datasource.go
```diff
@@ -142,13 +142,13 @@ func (plugin *GrafanaAzureDXDatasource) QueryData(ctx context.Context, req *back
 		return nil, err
 	}
 
-	email := "Noninteractive"
+	username := "Noninteractive"
```
kylebrandt, line 145: I'm not sure if it is more or less clean to move this into handleQuery, and pass plugincontext.user to the argument to handleQuery..

kylebrandt (approved):

pkg/datasource.go
sunker: Yeah I think so. I'll update. Thanks for feedback!

PR #202, opened 2021-01-18
Bugfix: Fix query variables 

joshhunt:
The `variables` member on the datasource class was conflicting with Grafana internals, causing query variables to not work.

This just renames the member to be a normal function and updates references.

It works again: !!!

image


sunker (approved):
LGTM

hugohaggmark (approved):
Great work ⭐ !

PR #215, opened 2021-03-17
Fix: Replace variables within template variable queries

### LINKED ISSUE #214: Issue with $__contains() in template variable queries
This seems to be related to #188.

I have various template variable queries which use the output from other template variables as input for where clauses. For these I use the $__contains() template macro for ease and efficiency.

For example, I have a query like this:

```
// template variable output_column
AggregationAll
| where $__contains(column_name1, $my_var1) 
| where $__contains(column_name2, $my_var2) 
| where $__contains(column_name3, $my_var3)
| where $__contains(column_name4, $my_var4) 
| distinct output_column
```

In ADX plugin 1.32 and 2.03 the following query is sent to ADX:

```
// template variable output_column
AggregationAll
| where 1 == 1 
| where 1 == 1 
| where 1 == 1
| where 1 == 1 
| distinct output_column
```
The above query is correct. 
However, when using ADX data source plugin 3.3.2 the following query is sent to ADX. 

```
// template variable output_column
AggregationAll
| where column_name1 in ('all') 
| where column_name2 in ('all') 
| where column_name3 in ('all')
| where column_name4 in ('all') 
| distinct output_column
```
As a result the query returns nothing.

We recently upgraded to from Grafana 6.7.4 to 7.4.3.

joshhunt:
Fixes #214

I created a `$states` template variable with the query `PopulationData | project State`, selected "Multi-value", "Include all option" and made the all value `all`.

Then I made a `$population` variable with the query

```
PopulationData
| where $__contains(State, $states)
| distinct Population
| limit 100
```

which was being sent over the wire (incorrectly) as 

```
PopulationData
| where State in ('all')
| distinct Population
| limit 101
```

After this change the query is sent as 

```
PopulationData
| where 1 == 1
| distinct Population
| limit 101
```

I was not able to get the second query to actually populate options for the variable for some reason, despite the same query producing results in a regular Grafana ADX panel.

### DIFF
--- a/src/datasource.ts
+++ b/src/datasource.ts
@@ -269,10 +269,11 @@ export class AdxDataSource extends DataSourceWithBackend<KustoQuery, AdxDataSour
       options = {};
     }
     if (!options.hasOwnProperty('scopedVars')) {
-      options['scopedVars'] = {};
+      options.scopedVars = {};
     }
 
-    const interpolatedQuery = interpolateKustoQuery(query, options['scopedVars']);
+    const replacedQuery = this.templateSrv.replace(query, options.scopedVars, this.interpolateVariable);
+    const interpolatedQuery = interpolateKustoQuery(replacedQuery, options.scopedVars);
 
     return {
       ...defaultQuery,


joshhunt:
So turns out there's an issue with ADX variable queries where it will only accept strings, so my test query that returned populations won't work.

I was able to validate this fix instead with the query

```
PopulationData
| where $__contains(State, $states)
| distinct State
| limit 100
```

and confirmed everything looks good!

sunker (approved):
LGTM

PR #216, opened 2021-03-17
ConfigEditor: Show errors from refreshing schema

joshhunt:
I missed this during converting the ConfigEditor to react

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "@grafana/data": "7.1.1",
     "@grafana/e2e": "7.1.1",
     "@grafana/eslint-config": "2.0.0",
-    "@grafana/runtime": "7.1.1",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -14,7 +14,7 @@
     "@grafana/data": "7.1.1",
     "@grafana/e2e": "7.1.1",
     "@grafana/eslint-config": "2.0.0",
-    "@grafana/runtime": "7.1.1",
+    "@grafana/runtime": "7.4.0",
```
sunker, line 17: Can you also update the dependencies in plugin.json? External plugins will receive packages from the grafana runtime, so if ADX is depending on grafana 7.4, this needs to be updated.

joshhunt: yes, i did :)

sunker (approved):
LGTM

PR #218, opened 2021-03-29
Fix: Clear empty WHERE rows

### LINKED ISSUE #121: When removing a where filter the whole line should be removed.
Happening: If we have multiple AND where filters and you remove the first of them the row is still rendered but with a [+] icon.
Expected: The whole row should be removed.

joshhunt:
Fixes #121

### DIFF
--- a/src/editor/components/QueryEditorRepeater.tsx
+++ b/src/editor/components/QueryEditorRepeater.tsx
@@ -22,7 +22,7 @@ export const QueryEditorRepeater: React.FC<Props> = props => {
   const onChange = useCallback(
     (index: number, expression?: QueryEditorExpression) => {
       const { expressions } = value;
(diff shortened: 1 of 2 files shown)


src/editor/components/QueryEditorRepeater.tsx
```diff
@@ -47,9 +51,9 @@ export const QueryEditorRepeater: React.FC<Props> = props => {
 
   return (
     <>
-      {value.expressions.map((val, idx) => {
+      {Array.from(value.expressions).map((val, idx) => {
         return (
-          <React.Fragment key={`${id}-${idx}`}>
```
sunker, line 52: Yes!!

sunker (approved):
Nice! Tested and it works as expected

Issue #220, opened 2021-04-05
Simplified query builder doesn't properly escape table names


mojanas:
Table names that have a . in them (ie `events.all`) need to be escaped when referenced in queries (`['events.all']`). The new simplified query builder doesn't do this, so the query fails with the error "Ensure that the expression is indeed a simple name."

![MicrosoftTeams-image (8)](https://user-images.githubusercontent.com/53838891/113594554-7b21a880-95fd-11eb-8488-440aacc4e0a2.png)

joshhunt:
Thanks for reporting!

joshhunt:
I wonder if fields on tables also exhibit this issue. We'll look into it.

PR #221, opened 2021-04-15
Format property names with special characters

sarahzinger:
Fix for: https://github.com/grafana/azure-data-explorer-datasource/issues/220

It looks like table names can contain dots, dashes, and spaces, and if they do they are required to be in [quotes](https://docs.microsoft.com/en-us/azure/data-explorer/kusto/query/schema-entities/entity-names#identifier-naming-rules).

As part of this work, we looked into column names containing dots, to see if we have a similar problem there. I don't think we want to change column names into quotes as we can have column names like `column.country` so if we force those into quotes we'll break existing functionality. I sort of suspect there must be some way to get back the type of these properties in the same way we do for the values, but I also haven't heard of users running into this problem, so I decided not to tackle that in this pr.

I spoke briefly with @mckn this morning and we think that tablenames that contain schema mappings will always contain either a `$` a `(` or both, ex: `StormEvents($__from, $__to)` so in those circumstances we do not wish to put the table name in quotes even if there's a space, dot, or dash.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## [3.3.3]
+
+- Bugfix: Fixed issue where query builder did not handle table names that contained special characters
+
 ## [3.3.2]
 
 - Bugfix: Fixed an issue where the KQL Monaco editor wouldn't load when Grafana is served from a sub path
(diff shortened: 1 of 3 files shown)


joshhunt (approved):
LGTM

PR #227, opened 2021-05-24
Refactoring: Instance manager, shared http client and token provider

### LINKED ISSUE #229: Use instance manager and have one http client per instance
At the moment, the ADX backend doesn't have instance management, so the HttpClient is recreated for each request to the backend. 

The backend should implement instance management and keep one HttpClient per instance. Then, auth needs to be refactored. Tokens should be retrieved and cached in a manner that is similar to [azure token provider](https://github.com/grafana/grafana/blob/main/pkg/api/pluginproxy/token_provider_azure.go) in core grafana.

sunker:
This refactoring is a first step laying ground for unitifed auth for all Azure data sources. Read more about that in [this](https://github.com/grafana/grafana/issues/33809) issue. Token logic and caching is now contained in a separate package, which will later be put in a shared lib that can be used by any Azure data source. 

What this PR does (besides from new token logic):
* Implement [automatic](https://github.com/grafana/grafana-plugin-sdk-go/blob/master/backend/datasource/manage.go) instance management. Before this refacgoring, the ADX data source did not have any instance management, so a new http client was created for each call to QueryData
*  Use new shared [http client](https://github.com/grafana/grafana-plugin-sdk-go/tree/master/backend/httpclient) from the go plugin sdk
*  Backend refactoring. Splitting up the code in different packages to increase readability and make the code easier to test (more tests will come)

@marefr would be great if I could get some feedback on how the SDK http client is implemented
@ryantxu would be great to get some feedback on how the automatic instance manager is implemented 

Fixes #229


pkg/azuredx/client/client.go
```diff
@@ -0,0 +1,120 @@
...
+	Log hclog.Logger
+}
+
+// NewClient creates a Grafana Plugin SDK Go Http Client
+func New(tokenProvider *tokenprovider.AccessTokenProvider) (*Client, error) {
+	c := Client{}
+	var err error
+
+	c.Client, err = sdkhttpclient.New(sdkhttpclient.Options{
+		Middlewares: []sdkhttpclient.Middleware{
```
sunker, line 32: This will give me all default values from the SDK Http Client, right? Should be no need for me to set time out etc?

pkg/azuredx/client/auth_middleware.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/grafana/azure-data-explorer-datasource/pkg/azuredx/tokenprovider"
+	"github.com/grafana/grafana-plugin-sdk-go/backend/httpclient"
+)
+
+const authenticationMiddlewareName = "Authentication"
+
+func authMiddleware(tokenProvider *tokenprovider.AccessTokenProvider) httpclient.Middleware {
+	return httpclient.NamedMiddlewareFunc(authenticationMiddlewareName, func(opts httpclient.Options, next http.RoundTripper) http.RoundTripper {
+		return httpclient.RoundTripperFunc(func(req *http.Request) (*http.Response, error) {
+			log.Println("Before outgoing request")
```
marefr, line 17: Use plugin logger with debug level

pkg/azuredx/client/auth_middleware.go
```diff
@@ -0,0 +1,29 @@
...
+			token, err := tokenProvider.GetAccessToken(req.Context())
+			if err != nil {
+				// next()
+			}
+			req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token))
+
+			res, err := next.RoundTrip(req)
+			log.Println("After outgoing request")
```
marefr, line 25: Use plugin logger with debug level

pkg/azuredx/datasource.go
```diff
@@ -1,30 +1,62 @@
...
+}
 
-func (plugin *GrafanaAzureDXDatasource) handleQuery(client *azuredx.Client, q backend.DataQuery, user *backend.User) backend.DataResponse {
+func NewDatasource(settings backend.DataSourceInstanceSettings) (instancemgmt.Instance, error) {
+	datasourceSettings := models.DatasourceSettings{}
+	err := datasourceSettings.Load(settings)
+	if err != nil {
+		return nil, err
+	}
+
```
marefr, line 29: You probable want to use `provider := httpclient.NewProvider(ProviderOptions)` here and add the auth token middleware through the provider options. Then you can put `provider` or `provider.New(settings.HTTPClientOptions())` as argument to `client.New`.

marefr (commented):
Looks good in general I think. Added some comments

Should the `pkg/__debug_bin ` file be commited?

pkg/azuredx/datasource.go
sunker: :+1:

sunker:
force pushed to remove pkg/__debug_bin

pkg/azuredx/tokenprovider/token_cache.go
```diff
@@ -0,0 +1,185 @@
...
+
+		if accessToken != nil {
+			c.accessToken = accessToken
+		}
+
+		c.cond.Broadcast()
+		c.cond.L.Unlock()
+	}()
+
+	token, err := c.credential.GetAccessToken(ctx, c.scopes)
```
andresmgot, line 168: if this is a simple Bearer token, why refreshing it? wouldn't you get the same token everytime? (or is this some oauth token?)

pkg/azuredx/client/client.go
```diff
@@ -0,0 +1,100 @@
...
+		errorData := &models.ErrorResponse{}
+		err = jsoniter.Unmarshal(bodyBytes, errorData)
+		if err != nil {
+			backend.Logger.Debug("failed to unmarshal error body from response", "error", err)
+		}
+		return nil, errorData.Error.Message, fmt.Errorf("HTTP error: %v - %v", resp.Status, bodyString)
+	}
+	tr, err := models.TableFromJSON(resp.Body)
+	return tr, "", err
+}
```
andresmgot, line 100: this file seems important. I know it's mostly doing requests to other service but can we cover it with unit tests?

andresmgot (commented):
The code looks good to me but you probably want a review from someone with more context :)

pkg/azuredx/client/auth_middleware.go
```diff
@@ -0,0 +1,34 @@
...
+const authenticationMiddlewareName = "Authentication"
+
+var logError = backend.Logger.Error
+
+type TokenProvider interface {
+	GetAccessToken(ctx context.Context) (string, error)
+}
+
+func AuthMiddleware(tokenProvider TokenProvider) httpclient.Middleware {
+	return httpclient.NamedMiddlewareFunc(authenticationMiddlewareName, func(opts httpclient.Options, next http.RoundTripper) http.RoundTripper {
```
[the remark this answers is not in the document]
sunker: good point!

pkg/azuredx/tokenprovider/token_cache.go
sunker: it's jwt token that expires after a certain time, so it needs to be refreshed

pkg/azuredx/client/client.go
sunker: yes good point, tests are coming up

sunker: added now!

sunker: I plan to add more client tests eventually, but I think the client might be a subject of refactoring when we add a ResourceHandler so I'll do that later

andresmgot (approved):
LGTM!

kostrse:
Looks good!

Some thoughts on future design when we move it into a shared library:
1. Rather than passing _tenantId_, _clientId_, _clientSecret_ and other fields individually, we'll need to introduce `AzureCredentails` structures, similar to what we have [in Azure Monitor](https://github.com/grafana/grafana/blob/20d356947c2bfaf461594bca451a43fe1b676a34/public/app/plugins/datasource/grafana-azure-monitor-datasource/types/index.ts#L57-L69). We could probably even serialize the whole structure as a single JSON object. Schema could be shared across plugins to reuse validation logic and stuff.
2. We'll need to decide on lifespan of token provider and whether it's going to keep any state in future. Right now a new instance being created for each request and all parameters being passed into it's constructor. Maybe it makes sense to preserve the instance between requests and pass request-specific parameters (credentials, scopes) into the `GetAccessToken` method.

sunker:
> 1. Rather than passing _tenantId_, _clientId_, _clientSecret_ and other fields individually, we'll need to introduce `AzureCredentails` structures, similar to what we have [in Azure Monitor](https://github.com/grafana/grafana/blob/20d356947c2bfaf461594bca451a43fe1b676a34/public/app/plugins/datasource/grafana-azure-monitor-datasource/types/index.ts#L57-L69). We could probably even serialize the whole structure as a single JSON object. Schema could be shared across plugins to reuse validation logic and stuff.

Completely agree! 

> 2. We'll need to decide on lifespan of token provider and whether it's going to keep any state in future. Right now a new instance being created for each request and all parameters being passed into it's constructor. Maybe it makes sense to preserve the instance between requests and pass request-specific parameters (credentials, scopes) into the `GetAccessToken` method.

With the changes in this PR, the life span of the token provider is equal to the life span of the instance. The instance will be preserved between requests - it will live until changes to the data source settings (i.e someone updates the data source in the config editor). For the token provider, I don't think any parameters should be request specific? They will all be unique per instance and datasource. 

I think the term `provider` suggests it's stateless. In reality, the token provider has a cache so it's not. For the shared package, I guess we can try to come up with a better name for this.

pkg/azuredx/datasource.go
```diff
@@ -1,30 +1,101 @@
...
+	httpClient, err := httpClientProvider.New(httpClientOptions)
+	if err != nil {
+		backend.Logger.Error("failed to create HTTP client", "error", err.Error())
+	}
+
+	return &GrafanaAzureDXDatasource{
+		client:   client.New(httpClient),
+		settings: &datasourceSettings,
+	}, nil
```
joshhunt, line 51: Not totally familiar with Go and it's practices, but should this return an error if it can't make an http client?

joshhunt (approved):
I can't comment too much on the Go code, but I have this a test and it seemed to work fine 👍🏻

pkg/azuredx/datasource.go
sunker: yes!!

PR #230, opened 2021-05-25
Stop throwing errors when url is not defined

sarahzinger:
Right now when you load the config editor for Azure Data Explorer you see 2 errors (one an orange-y toast, the other a static red error) that both show raw html. 

They don't block you from loading the datasource, but they are confusing.

I think there are a few things happening that we should fix:
1) we are making a request to update the schema without a url which is throwing an error. There's no need to make a request that we know will fail. (What this PR fixes)
2) props.options.url seems to always be an empty string. It's not clear to me what is supposed to set this or where these props are coming from exactly, but we should work on that. (Should fix!)
3) we are always setting these errors even when we get back html which seems unnecessary. Perhaps there is some way to only set the error if it's in a format we expect, or otherwise translate errors back to users in ways that are predictable. (This is optional to me but nice to fix if we can).

I'm not sure how to tackle 2 or 3 at the moment, but this pr felt like an improvement until we can figure out what's going on for 2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change Log
 
+## [next]
+- Bugfix: Fix an error loading schemas on the configuration page.
+
 ## [3.4.0]
 
 Note: The minimum required version of Grafana is now 7.4
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "@grafana/toolkit": "7.3.1",
     "@grafana/ui": "7.5.0-beta.1",
     "@testing-library/react": "^11.2.5",
+    "@testing-library/user-event": "^13.1.9",
     "@types/angular": "1.6.49",
     "@types/debounce-promise": "^3.1.2",
     "@types/grafana": "github:CorpGlory/types-grafana.git",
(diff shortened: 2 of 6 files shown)


andresmgot (approved):
a test would be nice for the fix :)

sunker (commented):
I think the problem is here in the refreshSchema.ts file
```typescript
  const response = await getBackendSrv().datasourceRequest({
    url: `${baseUrl}/azuredataexplorer/v1/rest/mgmt`,
    method: 'POST',
    data: data,
  });
```
The base url is not set correctly on the ds options, so therefore we're not calling the endpoint of the ADX backend api. Think a solution to that would be to use the getSchema method in datasource.ts. The datasource is not passed as a prop to the ConfigEditor, but you should be able to load it. It would be something like this:
`const datasource: AdxDataSource = await getDataSourceSrv().get(options.name);`

The data source doesn't exist the first time a user enters the config page, so we'll have to guard for that so we don't try to load schema then. But it should work for those users who already have a functioning configuration.

sarahzinger:
@sunker thanks for the suggestion, mind taking a look?

@andresmgot thanks for the suggestions! Will take a look at the prettier stuff in a future pr with the e2e testing work

After this is merged should I make a patch release since this is fairly noticeable?

joshhunt (approved):

joshhunt:
looks great - @sarahzinger lets merge this and release immediately as a patch release.

sunker (approved):
Awesome work @sarahzinger!

PR #231, opened 2021-05-27
Call resource handler

### LINKED ISSUE #232: Use CallResourceHandler instead of plugin proxy routes

sunker:
**What this PR does**
This PR replaces the usage of the plugin proxy in core grafana with a call resource handler in the adx backend. 

Before this PR, all client side requests went through the plugin proxy in grafana, receiving token based on credentials specified in the route in plugin.json. For reasons explained in [this](https://github.com/grafana/grafana/issues/33809) pr, we don't want to use the proxy in core grafana anymore.

All client side requests now go through the call resource handler that the backend implements. It's a basic http api the currently only has two routes.  The handler use the same http client as the QueryData method does, so auth will be handled in the same way for both types of requests. 

Fixes #232

**Notes to reviewer**
The client KustRequest could use some more refactoring, but I'd like to limit the amount of changes for now since there were no tests for this. 

Also note that the two routes, getSchema and getDatabases are very similar. It could have been just one handler for the management api, accepting a POST method with the payload in the body, but I think this is cleaner for now.

### DIFF
--- a/pkg/azuredx/datasource.go
+++ b/pkg/azuredx/datasource.go
@@ -60,6 +60,9 @@ func NewDatasource(settings backend.DataSourceInstanceSettings) (instancemgmt.In
 	adx.registerRoutes(mux)
 	adx.CallResourceHandler = httpadapter.New(mux)
 
+	return adx, nil
+}
+
 func (adx *AzureDataExplorer) Dispose() {
 	tokenCache.Purge()
 }
--- a/pkg/azuredx/datasource.go
+++ b/pkg/azuredx/datasource.go
@@ -2,32 +2,38 @@ package azuredx
 
 import (
 	"fmt"
+	"net/http"
 
+	"github.com/google/uuid"
 	"github.com/grafana/azure-data-explorer-datasource/pkg/azuredx/client"
 	"github.com/grafana/azure-data-explorer-datasource/pkg/azuredx/models"
 	"github.com/grafana/azure-data-explorer-datasource/pkg/azuredx/tokenprovider"
 	"github.com/grafana/grafana-plugin-sdk-go/backend"
 	sdkhttpclient "github.com/grafana/grafana-plugin-sdk-go/backend/httpclient"
 	"github.com/grafana/grafana-plugin-sdk-go/backend/instancemgmt"
+	"github.com/grafana/grafana-plugin-sdk-go/backend/resource/httpadapter"
 	"github.com/grafana/grafana-plugin-sdk-go/data"
 	jsoniter "github.com/json-iterator/go"
 	"golang.org/x/net/context"
 )
 
-// GrafanaAzureDXDatasource stores reference to plugin and logger
-type GrafanaAzureDXDatasource struct {
-	client   *client.Client
+// AzureDataExplorer stores reference to plugin and logger
+type AzureDataExplorer struct {
+	backend.CallResourceHandler
+	client   client.AdxClient
 	settings *models.DatasourceSettings
 }
 
 var tokenCache = tokenprovider.NewConcurrentTokenCache()
 
 func NewDatasource(settings backend.DataSourceInstanceSettings) (instancemgmt.Instance, error) {
-	datasourceSettings := models.DatasourceSettings{}
+	adx := &AzureDataExplorer{}
+	datasourceSettings := &models.DatasourceSettings{}
 	err := datasourceSettings.Load(settings)
 	if err != nil {
 		return nil, err
 	}
+	adx.settings = datasourceSettings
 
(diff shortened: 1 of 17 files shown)


marefr (approved):
Nice 👍

pkg/azuredx/datasource.go
```diff
@@ -124,12 +127,24 @@ func (plugin *GrafanaAzureDXDatasource) handleQuery(q backend.DataQuery, user *b
 		resp.Error = err
 	}
 
+	headers := map[string]string{}
+	msClientRequestIDHeader := fmt.Sprintf("KGC.%v;%v", qm.QuerySource, uuid.Must(uuid.NewRandom()).String())
```
andresmgot, line 131: `"KGC.%s;%s"` (these are strings, right?)

pkg/azuredx/datasource.go
```diff
@@ -124,12 +127,24 @@ func (plugin *GrafanaAzureDXDatasource) handleQuery(q backend.DataQuery, user *b
...
+	}
+	headers["x-ms-client-request-id"] = msClientRequestIDHeader
+
 	var tableRes *models.TableResponse
-	tableRes, kustoError, err = plugin.client.KustoRequest(plugin.settings, models.RequestPayload{
-		CSL:        qm.Query,
-		DB:         qm.Database,
-		Properties: models.NewConnectionProperties(plugin.settings, cs),
-	}, qm.QuerySource, user)
+	tableRes, _, kustoError, err = adx.client.KustoRequest(adx.settings.ClusterURL+"/v1/rest/query", models.RequestPayload{
```
andresmgot, line 141: I see that we are concatenating urls with `+` everywhere but for future reference `ResolveReference` should be a more robust solution: https://golang.org/pkg/net/url/#URL.ResolveReference (it would avoid, for example, double slashes)

andresmgot (approved):
Looks great! LGTM

joshhunt (approved):

joshhunt:
oops i didnt mean to approve, just comment. giving this a test now

joshhunt:
Hmmm am I missing something - the query editor fails to load. It gets a 404 on `/api/datasources/1375/resources/schema`.

I've rebuilt both the ADX plugin and grafana a few times and I can't seem to get it to work?

image

pkg/azuredx/datasource.go
sunker: yes good idea! we're gonna do lots of refactoring to the adx backend in the next few months. this is certainly one of those things!

src/datasource.ts
```diff
@@ -189,17 +186,7 @@ export class AdxDataSource extends DataSourceWithBackend<KustoQuery, AdxDataSour
...
-        const req = {
-          querySource: 'schema',
-          csl: `.show databases schema as json`,
-        };
-
-        return this.doRequest(url, req).then(response => {
-          return new ResponseParser().parseSchemaResult(response.data);
-        });
-      },
+      () => this.getResource('schema').then(new ResponseParser().parseSchemaResult),
```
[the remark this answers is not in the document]
sunker: 404! the SDK takes care of that

sunker:
> Hmmm am I missing something - the query editor fails to load. It gets a 404 on `/api/datasources/1375/resources/schema`.
> 
> I've rebuilt both the ADX plugin and grafana a few times and I can't seem to get it to work?

Hmm interesting. The plugin is refactored to use automatic instance management, and with that the plugin is served using a different signature. That comes with backend plugin debugging support in vscode, but the downside of that might be you actually HAVE to start the debugger for it to work. Can you try using the following steps? 

1. `mage buildAll` in the plugin repo
2. `yarn dev` in the plugin repo
3. `make run` in grafana
4. F5 to start the debugger in the plugin

Please note that I had to rebase master

sunker:
Also deployed to our azure test env, so you should be able to test it there

joshhunt:
Oh - I was running `make run` instead of `mage buildAll`. Silly me.

All works fine 👍🏻

andresmgot (approved):
Still LGTM. Just one minor comment.

pkg/azuredx/resource_handler.go
```diff
@@ -0,0 +1,77 @@
...
+		respondWithError(rw, statusCode, kustoErr, err)
+		return
+	}
+
+	rw.Header().Set("Content-Type", "application/json")
+	err = json.NewEncoder(rw).Encode(response)
+	if err != nil {
+		rw.WriteHeader(http.StatusInternalServerError)
+	}
+}
```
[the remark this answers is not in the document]
sunker: correct! first I had just one handler for the two different routes, but using different payloads. but as mentioned in the description, I though this was cleaner even though it duplicated the code.

PR #233, opened 2021-05-28
Purge token cache before receiving new ds settings

sunker:
The token cache life span is currently equal to the lifespan of the backend. However, the cache was never purged. This PR purges the cache before the datasource receives new settings.

Token provider tests are coming in a separate PR for handling national clouds.


andresmgot (approved):
:+1:

PR #234, opened 2021-05-28
Add national clouds support

### LINKED ISSUE #16: Support Azure National Clouds
I noticed that AAD endpoint currently is hard coded to Azure public cloud,
and according to https://docs.microsoft.com/en-us/azure/active-directory/develop/authentication-national-cloud

National cloud | Azure AD authentication endpoint
-- | --
Azure AD for US Government | https://login.microsoftonline.us
Azure AD Germany | https://login.microsoftonline.de
Azure AD China operated by 21Vianet | https://login.chinacloudapi.cn
Azure AD (global service) | https://login.microsoftonline.com

I wish we can add an extra config for AAD endpoint (a dropdown or simple input) to support national clouds:)

Actually I made it locally, but I have no idea how to build the code (not familiar with js stuff), or I can try create a PR for you:)

sunker:
Adding support for national clouds. Been trying to keep names aligned with what we have in Azure Monitor. Will make things easier when we break these things out to a separate go and npm package. 

Will change merge base as soon as the call-resource-handler branch is merged.

Fixes #16
Closes #217

### DIFF
--- a/pkg/azuredx/datasource.go
+++ b/pkg/azuredx/datasource.go
@@ -25,6 +25,7 @@ type AzureDataExplorer struct {
 }
 
 var tokenCache = tokenprovider.NewConcurrentTokenCache()
+const AdxScope = "https://kusto.kusto.windows.net/.default"
 
 func NewDatasource(settings backend.DataSourceInstanceSettings) (instancemgmt.Instance, error) {
 	adx := &AzureDataExplorer{}
@@ -35,7 +36,7 @@ func NewDatasource(settings backend.DataSourceInstanceSettings) (instancemgmt.In
 	}
 	adx.settings = datasourceSettings
 
-	tokenProvider := tokenprovider.NewAccessTokenProvider(tokenCache, datasourceSettings.ClientID, datasourceSettings.TenantID, "AzurePublic", datasourceSettings.Secret, []string{"https://kusto.kusto.windows.net/.default"})
+	tokenProvider := tokenprovider.NewAccessTokenProvider(tokenCache, datasourceSettings.ClientID, datasourceSettings.TenantID, datasourceSettings.AzureCloud, datasourceSettings.Secret, []string{"https://kusto.kusto.windows.net/.default"})
 
 	httpClientProvider := sdkhttpclient.NewProvider(sdkhttpclient.ProviderOptions{
 		Middlewares: []sdkhttpclient.Middleware{
--- a/pkg/azuredx/tokenprovider/clouds.go
+++ b/pkg/azuredx/tokenprovider/clouds.go
@@ -1,0 +1,8 @@
+package tokenprovider
+
+// Must be in synch with azure cloud options in ConfigEditor.tsx
+const (
+	AzurePublic       = "azuremonitor"
+	AzureChina        = "chinaazuremonitor"
+	AzureUSGovernment = "govazuremonitor"
+)
--- a/pkg/azuredx/tokenprovider/token_provider_test.go
+++ b/pkg/azuredx/tokenprovider/token_provider_test.go
@@ -1,0 +1,116 @@
+package tokenprovider
+
+import (
+	"context"
+	"testing"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/stretchr/testify/require"
+)
+
+const TestClientId = "3c882424-c381-45c0-a123-0a62ebc989dc"
+const TestTenantId = "b931e0dd-f42c-444a-83a9-63de6f2d771a"
+const TestClientSecret = "3093e953-91e1-48f4-9ad0-e8f750c38f93"
+const Scope = "https://kusto.kusto.windows.net/.default"
+const ExpectedToken string = "4cb83b87-0ffb-4abd-82f6-48a8c08afc53"
+
+type tokenCacheFake struct{}
+
+func (c *tokenCacheFake) GetAccessToken(_ context.Context, credential TokenCredential, scopes []string) (string, error) {
+	credential.Init()
+	return ExpectedToken, nil
+}
+
+func (c *tokenCacheFake) Purge() {
+	panic("not implemented")
+}
+
+func TestTokenProvider(t *testing.T) {
+	ctx := context.Background()
+
+	newClientSecretCredentialOriginal := newClientSecretCredential
+
+	tokenCacheFake := &tokenCacheFake{}
+
+	t.Cleanup(func() {
+		newClientSecretCredential = newClientSecretCredentialOriginal
+	})
+
+	t.Run("should resolve public cloud host when nothing is specified", func(t *testing.T) {
+		newClientSecretCredential = func(tenantID string, clientID string, clientSecret string, options *azidentity.ClientSecretCredentialOptions) (*azidentity.ClientSecretCredential, error) {
+			require.Equal(t, TestClientId, clientID)
+			require.Equal(t, TestTenantId, tenantID)
+			require.Equal(t, TestClientSecret, clientSecret)
+			require.Equal(t, azidentity.AzurePublicCloud, options.AuthorityHost)
+
+			return &azidentity.ClientSecretCredential{}, nil
+		}
+
+		// getAccessToken = func(ctx context.Context, credential TokenCredential, scopes []string) (string, error) {
+		// 	credential.Init()
+		// 	return ExpectedToken, nil
+		// }
+		provider := NewAccessTokenProvider(tokenCacheFake, TestClientId, TestTenantId, "", TestClientSecret, []string{Scope})
+		token, err := provider.GetAccessToken(ctx)
+		require.Nil(t, err)
+		require.Equal(t, ExpectedToken, token)
+	})
+
+	t.Run("should resolve public cloud host when public cloud is specified", func(t *testing.T) {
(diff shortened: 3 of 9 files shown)


src/components/ConfigEditor/ConnectionConfig.tsx
```diff
@@ -19,6 +24,13 @@ const ConnectionConfig: React.FC<ConnectionConfigProps> = ({
 }) => {
   const { jsonData, secureJsonData, secureJsonFields } = options;
 
+  // Set some default values
+  useEffect(() => {
+    if (!jsonData.azureCloud) {
+      updateJsonData('azureCloud', azureClouds[0].value);
```
joshhunt, line 30: I think it would be nice to have these clouds as some stable consts or enums, or from a json file we share with the backend maybe? 👀

I'm just trying to think about how we can eliminate problems with typos and things like that.

pkg/azuredx/datasource.go
```diff
@@ -35,7 +36,7 @@ func NewDatasource(settings backend.DataSourceInstanceSettings) (instancemgmt.In
 	}
 	adx.settings = datasourceSettings
 
-	tokenProvider := tokenprovider.NewAccessTokenProvider(tokenCache, datasourceSettings.ClientID, datasourceSettings.TenantID, "AzurePublic", datasourceSettings.Secret, []string{"https://kusto.kusto.windows.net/.default"})
+	tokenProvider := tokenprovider.NewAccessTokenProvider(tokenCache, datasourceSettings.ClientID, datasourceSettings.TenantID, datasourceSettings.AzureCloud, datasourceSettings.Secret, []string{"https://kusto.kusto.windows.net/.default"})
```
joshhunt, line 39: I'm not sure I see if a default value being set for the cloud in the backend?

`resolveAuthorityHost` has a fallback value, but I wonder if we should set a default up here at the top so elsewhere in the code we know that the value is always set (like for the [cache key](https://github.com/grafana/azure-data-explorer-datasource/blob/074ff5cc873f98e6713e7cb873fdd242fa0537f5/pkg/azuredx/tokenprovider/token_provider.go#L78)?)

joshhunt (commented):
Haven't tested it yet, just looked over the code.

I wonder if it would be smart to use the same config between frontend and backend for the cloud values?

pkg/azuredx/tokenprovider/token_provider_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/stretchr/testify/require"
+)
+
+const TestClientId = "3c882424-c381-45c0-a123-0a62ebc989dc"
+const TestTenantId = "b931e0dd-f42c-444a-83a9-63de6f2d771a"
+const TestClientSecret = "3093e953-91e1-48f4-9ad0-e8f750c38f93"
+const Scope = "https://kusto.kusto.windows.net/.default"
+const ExpectedToken string = "4cb83b87-0ffb-4abd-82f6-48a8c08afc53"
```
sarahzinger, line 15: just double checking we're cool with having all of these ids/secrets/etc be public? I guess I would think most tests would be with mocks, so it wouldn't matter what ids we used, but maybe that's not true.

src/datasource.ts
```diff
@@ -36,7 +36,7 @@ export class AdxDataSource extends DataSourceWithBackend<KustoQuery, AdxDataSour
   private backendSrv: BackendSrv;
   private templateSrv: TemplateSrv;
   private defaultOrFirstDatabase: string;
-  url?: string;
+  private url?: string;
```
sarahzinger, line 39: I think we want to keep this public because we use it here https://github.com/grafana/azure-data-explorer-datasource/blob/master/src/components/ConfigEditor/index.tsx#L43

sarahzinger (commented):
🎉

pkg/azuredx/tokenprovider/clouds.go
```diff
@@ -0,0 +1,8 @@
+package tokenprovider
+
+// Must be in synch with azure cloud options in ConfigEditor.tsx
+const (
+	AzurePublic       = "azuremonitor"
+	AzureChina        = "chinaazuremonitor"
+	AzureUSGovernment = "govazuremonitor"
+)
```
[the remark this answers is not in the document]
sunker: It is feasible..but very soon, like in a week or two this will be handled in the azure-sdk repo. I think we said last time that it will be two repos - one for go and one for typescript, so in that case we won't be able to share files like that. I think having that possibility would be nice though. Maybe we should reconsider having one repo for everything. But for now, let's keep it as it is?

src/components/ConfigEditor/ConnectionConfig.tsx
```diff
@@ -1,9 +1,14 @@
...
+import { FieldSet, InlineField, Input, LegacyForms, Select } from '@grafana/ui';
+import React, { useEffect } from 'react';
 import { AdxDataSourceOptions, AdxDataSourceSecureOptions } from 'types';
 
 const { SecretFormField } = LegacyForms;
+const azureClouds = [
+  { value: 'azuremonitor', label: 'Azure' },
+  { value: 'govazuremonitor', label: 'Azure US Government' },
+  { value: 'chinaazuremonitor', label: 'Azure China' },
+] as SelectableValue[];
```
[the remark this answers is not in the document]
sunker: Yeah yeah of course. I copied this from Azure Monitor without using my brain.

src/components/ConfigEditor/ConnectionConfig.tsx
sunker: I agree!

pkg/azuredx/datasource.go
sunker: I actually think it's sufficient to handle this is in the token provider. The tokenprovider package is meant to become the azure-sdk-go, and I'd like to push guards like this into the SDK to protect from future plugin developers from making mistakes. WDYT?

pkg/azuredx/tokenprovider/token_provider_test.go
sunker: They're not actual credentials, I just generated the GUIDs online so that the have the format of credentials so it should be fine. But thanks for checking!

src/datasource.ts
sunker: I actually had to refactor that line of code in [this](https://github.com/grafana/azure-data-explorer-datasource/pull/231/files) PR, so it should be fine I think.

pkg/azuredx/tokenprovider/token_provider_test.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func (c *tokenCacheFake) Purge() {
+	panic("not implemented")
+}
+
+func TestTokenProvider(t *testing.T) {
+	ctx := context.Background()
+
+	newClientSecretCredentialOriginal := newClientSecretCredential
```
andresmgot, line 34: Having a package variable for the test is a bit risky I would move it to an interface and either pass it as an argument in the `Init` method or store it in the `clientSecretCredential` but up to you if you don't think it's worth it.

src/components/ConfigEditor/ConnectionConfig.tsx
```diff
@@ -47,6 +59,18 @@ const ConnectionConfig: React.FC<ConnectionConfigProps> = ({
...
     <FieldSet label="Connection Details">
+      <InlineField label="Azure cloud" labelWidth={26} tooltip="Select an Azure Cloud.">
+        <Select
+          options={azureClouds}
+          value={azureClouds.find(v => v.value === jsonData.azureCloud)}
+          onChange={(change: SelectableValue<AzureCloudType>) =>
+            updateJsonData('azureCloud', change.value ? change.value : AzureCloudType.AzurePublic)
+          }
+          isClearable={false}
+          width={60}
```
andresmgot, line 70: I guess we can add the `required` property here so it's not empty? (even though I think it cannot be empty with the default value)

andresmgot (approved):
LGTM, some minor comments

pkg/azuredx/datasource.go
joshhunt: Yeah that makes sense. Just as long as value is consistent.

pkg/azuredx/tokenprovider/clouds.go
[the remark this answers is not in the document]
joshhunt: Fair point. Another option, for later, could be for the client to request the list from the backend through a resource handler? idk.

joshhunt (approved):
Gave it a test and its looking good! 👍🏻

pkg/azuredx/tokenprovider/clouds.go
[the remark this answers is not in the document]
sunker: Would be nice, but unfortunately it wouldn't work the first time you enter the config page since there's no backend created for it yet.

pkg/azuredx/tokenprovider/token_provider_test.go
sunker: yeah that's true. hmm at the moment, the token cache is generic and I'd like to avoid adding a dependency to azidentity there. think I'll need to keep it as is for now and revisit this later when there's more time. thanks for good feedback!

floushee:
As this was part of todays 3.5.0 release I wanted to give it a go and it looks like authentication is not working against US Government Azure Cloud. I have created an issue for this: https://github.com/grafana/azure-data-explorer-datasource/issues/235

Issue #237, opened 2021-06-02
Config Input for Client Secret loses focus after typing a character


sarahzinger:
To recreate:
- add a new azure data explorer datasource
- in the config editor click on the input for client secret
- start typing the client secret (do not paste in a secret), notice that the onblur fires off after the first character is typed.

PR #240, opened 2021-06-16
Adding e2e tests to adx

sarahzinger:
## Adding e2e test flow for testing datasources
The goal of this pr was to create a function `smokeTestDatasource` which other datasources could use that could quickly run through a few generic e2e tests. The thinking was this would be easier to explore with a specific datasource so we started with adx.

Some reasons to have some e2e tests for our datasource plugins:
- Each plugin relies on interfaces from both core grafana and from their cloud providers that ideally shouldn't change much but could. While we could use mocks for both core grafana and for our cloud provider's apis, mocks will not help us monitor if unexpected breaking changes have come up.
- Ideally our plugins can be used on a wide range (or at least a known range) of different versions of grafana, however right now testing this is a difficult, manual, time consuming process (that is often not done). Once we have e2e tests we like, we can try running them against different grafana versions with some kind of script. 
- as the cloud datasources team, we maintain access to several testing environments across all of our different cloud providers. Overtime, credentials expire, databases stop ingesting data, often without anyone noticing until a while later, when it is more difficult to resolve. Having e2e tests can help us alert us to expired test environments, and can point to at least a portion in time when there was up to date test data. 

This PR touches a TON of files (my apologies!), mostly because in order to use some of more recent e2e testing framework stuff, I needed to upgrade a few packages, which used some different prettier/linting rules. I tried to break that work up into separate commits so it's easier to look at. 

Here 2 files related to e2e testing to definitely checkout:
- `cypress/setup.ts` 
the concept here is that this file would live in grafana/e2e eventually and shouldn't really know specifics about any particular datasource but instead will work for all of them. You simply invoke it with the name of the datasource, how to set the credentials for it, the dashboard to smoke test wtih, and how you think that the formatted data should come back for the first panel in that dashboard. It focuses on testing that formatted dashboard object instead of focusing on screenshots, with the thinking that while our UI for how to render things might change overtime, the querying of that data and then formatting that data for a panel hopefully doesn't change much without us knowing about it.
- `spoke.spec.ts`
This is hopefully the kind of test file that would live in this repo. You'll notice I make a call to a local copy of the provisioning repo for both the datasource and for the dashboard. Unfortunately in the process of making this PR our credentials for that datasource expired (along with the dashboard that corresponds to  it). For the purposes of getting eyes on this PR sooner, I just created my own dashboard that shows a simple table. Perhaps before merging though I could get the new credentials for that old datasource instead.

### How to test locally
First make sure to run yarn install as I've upgraded several packages. 

Next, these e2e tests, similar to other e2e tests elsewhere, rely on a private repo called `plugin-provisioning` that contains various information necessary to connect to our test datasources. The current recommendation is to use symlinks so for example, from within the root of the adx repo I ran:
```
ln -s ../../plugin-provisioning/provisioning
```
which connected to my local version of the provisioning directory. 

From there you should be able to run:
```
yarn run e2e:open
```
and click to start the test in cypress

### Things I could use help with and feedback on
- *Coupling of expectedData and provisioning*: one thing I don't really love is that our smoke tests both require us to pass in provisioning information as well as a json object representing how we expect that data to eventually be formatted (called `expectedData`). And yet the provisioning information lives in one repo, and this expectedData object lives in another, despite being deeply coupled. I could try to add expectedData to the plugin-provisioning repo if we think that's a good approach maybe? Or maybe there's another solution to this that we like better that I can't think of at the moment? It feels a bit wrong to me as is though. 
- `Why is the build failing?` I seem to be getting some errors about node versions? I'm sort of assuming this has something to do with me upgrading toolkit, but I'm not really sure. If anyone has ideas on this that would be great! I do get `Using Node.js [object Object]` locally , which is the same error I see in the build in ci, although the build doesn't fail locally. 
- `What to do about .visit not working properly?` You'll notice in setup.ts I have a line that references localhost. Cypress supports relative links, you just have to set a baseUrl in the initial configuration. My expectation was that the e2e package would have made a default assumption for me that I'm running on localhost:3000 or would have given me a configuration setting to explicitly set that. I could be wrong but I don't think that's happening? I see there's a cool function `fromBaseUrl` but it's not exported as part of the package I don't think? I sort of suspect I'm missing a helper somewhere, but if not I could probably address that when I put this in the main grafana repo. 
- `Why are we removing node modules in one of our npm scripts` One of the more challenging aspects of this PR has been upgrading various dependencies and then dealing with the typescript errors that came up in the process. I got a fair amount of linting/prettier/typescript errors, some legit and some not so much. One of the main ones that I kept seeing that seemed challenging (and possibly wrong?) to fix was related to a difference between `DataSourceApi` and `DataSourceWithBackend`, which is some sort of rxjs type conflict. For reasons that are not 100% clear to me both grafana/data and grafana/toolkit install different versions of rxjs which leads to this typescript error. The current hack that is being used to handle this challenge within the timestream datasource is to remove the duplicate dependency in an [npm script](https://github.com/grafana/timestream-datasource/blob/main/package.json#L9), so that's what I've done here. I don't love this solution, but I'm letting it go for now, to tackle in another pr, unless we have easy to implement other ideas!
- `Should this be part of our ci process` My thinking here is no. For one it's slow, two this really shouldn't change much, and just because a test environment's credentials expire shouldn't lead to a failing build that prevents deploys. My thinking is that this could run once a day or something in a github action. We could start with it running as is, but maybe could also attempt to think about how to use this e2e test to test against different versions of grafana as well. Perhaps if the tests don't pass we automatically open an issue for someone to debug. 
- `External contributors` I don't love that to run the e2e tests you kind of need to be internal to grafana or share the test credentials, this kind of goes back to the tightly coupled thing I mentioned at the beginning of this very long pr description between the plugin-provisioning repo and this expectedData bit. I'm not entirely sure what can be done about this except write up some kind of documentation around it. Maybe in an ideal world we have nice documentation that walks you through how to create these test databases with our cloud providers and feed it some data to ingest, idk probably beyond the scope of this PR. But open to other ideas if we have any!

### Next steps:
- I think first lets make sure this approach is agreeable to folks, as it's possible everyone hates this thing and I can just close it lol. 
- Then let's make sure this PR passes the build and merge it
- Then I can take some things from this pr and put them in the e2e package in grafana
- Then I can swap out what we have here for what's in there when I upgrade to the latest version of the e2e package. 

Addresses: https://github.com/grafana/grafana/issues/33524
Incidentally Fixes https://github.com/grafana/azure-data-explorer-datasource/issues/237 (this just came up in the process of testing)


sunker:
Thanks @sarahzinger! 

@vickyyyyyyy maybe you can chime in on this one?

vickyyyyyyy:
This is 🔥 
Super excited to see better test coverage 💯 

> we can try running them against different grafana versions with some kind of script

@toddtreece has been working a lot with the drone builds to get our enterprise plugins running against different Grafana versions (to check backwards compatibility) with cron jobs for frequent, scheduled builds. Could this work be applied to this or would there be more things to set up?

> shouldn't really know specifics about any particular datasource but instead will work for all of them

This is interesting, I was wondering how this would work for different datasources with different fields/UI for the query editors?

> It focuses on testing that formatted dashboard object instead of focusing on screenshots

I agree, the screenshot comparisons can be very flaky. We are currently using screenshot comparisons to test the following:
- The data comes back the same (this is tricky because the data we get from the API may be unstable due to data retention periods but this would be fixed with possibly seeding the data)
- Annotations are working
- Template variables are interpolated correctly

We haven't figured out a better way of testing these without screenshot comparisons - would be good if anyone had any other suggestions.

Testing the formatted dashboard object is a great idea, but I wonder if it makes sense for that to be more of an integration test rather than an e2e test? What do you think?

> in order to use some of more recent e2e testing framework, I needed to upgrade a few packages, which had some different prettier/linting rules

Interesting, does this mean the `@grafana/e2e` and `@grafana/e2e-selectors` packages are inconsistent with the ADX plugin?

> And yet the provisioning information lives in one repo, and this expectedData object lives in another, despite being deeply coupled

I'm not sure I'm following, why they are these deeply coupled?

> I do get Using Node.js [object Object] locally although the build doesn't fail locally

What version of node are you using?

> What to do about .visit not working properly?

What's the problem with this?

> a difference between DataSourceApi and DataSourceWithBackend, which is some sort of rxjs type conflict

Not sure if this helps, but we have recently replaced our removal of duplicate deps with resolving to a specific version of `rxjs` e.g. https://github.com/grafana/appdynamics-grafana-datasource/blob/master/package.json#L41

```
"resolutions": {  "rxjs": "6.6.3"  }
```

> Should this be part of our ci process

Agreed with your points, hopefully the scheduled stuff @toddtreece has been working on should help with enabling these to run in the background

> I don't love that to run the e2e tests you kind of need to be internal to grafana or share the test credentials

Yeah, this was something we were thinking about too with running the e2e tests for the Github plugin (which is also OSS): https://github.com/grafana/github-datasource because we needed to think about how we could run the e2e tests in the build without sharing our credentials. One way we were thinking we could do this was by using read-only access credentials that could be shared publicly. For Github, this was fine because there weren't many ways this could be exploited but I'm not sure if this could work with ADX?

But at least this could be run locally as long as the user uses their own credentials in the local provisioning file

> Maybe in an ideal world we have nice documentation that walks you through how to create these test databases and feed it some data to ingest

Maybe we could automate something to seed databases with data we expect? Maybe we could have some test fixtures that define what data we want and then this could help with knowing what our expected data would be?

sarahzinger:
Hi @vickyyyyyyy, thanks for the feedback! Maybe we can discuss synchronously on this at some point (and with Todd as well) but for now here's some thoughts to some of your questions:

**How would this work for different datasources if the ui is different fields/query editors?**
So my proposal is that e2e would have this [smokeTestDatasource](https://github.com/grafana/azure-data-explorer-datasource/pull/240) function which takes in a sample dashboard that would be unique to each datasource and contains whatever query that makes sense for that datasource, as well as whatever expectedData we'd expect that dashboard to generate for a panel (that again would be unique for that datasource). Does that make sense?

**Screenshots and Integration tests**
I think my general impression is that datasource plugins don't need to concern themselves too much with the ui of panels and things. While every datasource plugin has a UI, it kind of feels to me like the bulk of the logic is making requests to a cloud datasource like azure, then taking that data and formatting it so that core grafana components can process it. So in that sense I'm not sure if snapshot testing is totally necessary to me, but I could be persuaded to try them if we think it's worth it! I think another reason I kind of hesitated on snapshot testing and went with directly comparing objects instead was because I was concerned about the fact that we show non-deterministic things in our UI (such as how long the processing time would be to make the request) and it seemed easy to delete those properties in a test. To the question of why not make this an integration test, I think this basically is an integration test that is made easier with cypress tooling. In fact it's in an integration folder in the cypress folder haha. I think it's using our e2e testing library to ensure our datasource plugin is integrating into grafana as a whole. But maybe you have another suggestion in mind! I guess the world of testing is more of a spectrum.

**package management stuff**
I'm not sure I'd say that e2e and e2e-selectors are inconsistent with adx as much as I'd say that upgrading them threw the rxjs + prettier errors so I figured better just update everything anyway.
Also that resolutions property works too, just added it. I still think it's weird we need this but a problem for another day :)

**Deeply Coupled thing**
So [expectedData.ts](https://github.com/grafana/azure-data-explorer-datasource/pull/240/files#diff-f9a7fe32ba7156948a0946d0cf4da5b4be6d3a043227c8a8c07c1dddff3a3311R19) contains information about a query string that came from a dashboard that will normally live in the provisioning repo. So the question is do we want to maybe have these 2 things in the same repo? Does that make sense? 

**What version of node am I using?**
v14.16.1

**What's the problem with .visit**
If I use .visit with a relative path, I get the following error: `Sorry, we could not load: /dashboard/import 404 - Not Found (text/html)` I believe this is because it doesn't have a baseUrl of localhost:3000, so I pass it localhost manually in the code which resolves the issue, but that doesn't feel right. I think it should like know a default config or something?

sarahzinger:
Ok going to close this, but use this as a baseline for a few other prs.

PR #243, opened 2021-06-28
Bugfix: Fixed issue when typing vs copy/pasting client secret in config

sarahzinger:
Fixes: https://github.com/grafana/azure-data-explorer-datasource/issues/237

To reproduce on master branch:
trying typing in client secret instead of copying and pasting a client secret, you'll notice it loses focus when it should not.

### DIFF
--- a/src/components/ConfigEditor/ConnectionConfig.tsx
+++ b/src/components/ConfigEditor/ConnectionConfig.tsx
@@ -139,7 +139,7 @@ const ConnectionConfig: React.FC<ConnectionConfigProps> = ({
         inputWidth={30}
         placeholder=""
         onReset={() => handleClearClientSecret()}
-        onChange={handleClientSecretChange}
+        onBlur={handleClientSecretChange}
         isConfigured={!!secureJsonData?.clientSecret || !!secureJsonFields?.clientSecret}
         tooltip={clientSecretTooltip}
       />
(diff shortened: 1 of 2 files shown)


src/components/ConfigEditor/ConnectionConfig.tsx
```diff
@@ -139,7 +139,7 @@ const ConnectionConfig: React.FC<ConnectionConfigProps> = ({
         inputWidth={30}
         placeholder=""
         onReset={() => handleClearClientSecret()}
-        onChange={handleClientSecretChange}
+        onBlur={handleClientSecretChange}
         isConfigured={!!secureJsonData?.clientSecret || !!secureJsonFields?.clientSecret}
```
andresmgot, line 143: The problem with this fix is that if the user clicks outside the input while they are still typing it will hit the bug anyway. What I would do is change the condition `isConfigured` to only be true if the `clientSecret` has some value when first loading the component.

(another reason to keep working on https://github.com/grafana/grafana/issues/33809 since this bug is not present for Azure Monitor)

sarahzinger:
@andresmgot great suggestion thanks! Want to take another look for me? I notice I'm having issues with the reload schema button, but I'm thinking that's an unrelated bug.

andresmgot (approved):
:+1:

PR #254, opened 2021-07-28
Eliminate Client ID Panic

pascaldekloe:
The cryptographic-secure constraint from UUID v4 seems unnecessary here.


andresmgot (commented):
Hi @pascaldekloe, on a first look, this seems like a regression? The `uuid` has a bigger entropy and follows a well-known format for a unique ID as requested here: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-ncnbi/817da997-30d2-4cd3-972f-a0073e4e98f7

You mentioned in the title that this fixes a panic? Can you give more details about that? (maybe there is some other way to fix it).

pascaldekloe:
No regression @andresmgot. The UUID specification requires a cryptographic random source, which causes error scenario. In Go, all functions starting with `Must...` panic on error instead.

For a client trace 18446744073709551615 options should be more than plenty I'd say. 😂

What would be the advantage of an UUID over this even more well known format: "hexadecimal"? No need for dependencies to get a random token (since the 80s).

andresmgot:
Gotcha, probably I'm being too protective but since I don't know the original reason to set that as a UUID (or if some tool already assumes that format) I prefer to play it safe :) What if we avoid that `Must...`? e.g.

```
func getUUID() string {
  uuid, err := uuid.NewRandom()
  if err != nil {
    // Fallback to a random number
    return string(rand.Uint64())
  }
  return uuid.String()
}
```

pascaldekloe:
The header is supposed to contain a token of unspecified format. I'd make a change exactly to ensure that no tool assumes a format. 🙃

andresmgot:
well it's not explicitly set in the documentation so I don't feel strongly about this topic. Can you fix the conflict?

PR #263, opened 2021-09-17
Fix HTTP timeout setting not being used

### LINKED ISSUE #262: The Datasource Timeout property isn'
**What happened**:

**Screenshots**

**Environment**:
- Grafana version:
- Plugin version:
- OS Grafana is installed on:
- User OS & Browser:
- Others:

joshhunt:
Fixes #262

Tested by setting the ADX timeout in settings to 999s and using the following query which (aside from probably setting the cluster on fire...) takes roughly 3.5 minutes to run

```
range N from 1 to 70000 step 1
| extend Sub = range(1, N)
| mv-apply X=Sub on (
    where N % X == 0
    | summarize divisors=make_list(X)
  )
| where array_length(divisors) == 2
| project Prime=N
```

image

I do still want to test to make sure that the default timeout of 30s is still applied if the user has not specified a timeout.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,9 @@
 # Change Log
 
 ## [next]
+
 - Bugfix: Fixed issue when typing vs copy/pasting client secret in configuration
+- Bugfix: Fixed issue where HTTP timeout setting was not being applied
 
 ## [3.5.0]
 
--- a/pkg/azuredx/datasource.go
+++ b/pkg/azuredx/datasource.go
@@ -13,6 +13,7 @@ import (
 	"github.com/grafana/grafana-plugin-sdk-go/backend/instancemgmt"
 	"github.com/grafana/grafana-plugin-sdk-go/backend/resource/httpadapter"
 	"github.com/grafana/grafana-plugin-sdk-go/data"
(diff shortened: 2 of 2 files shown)


sunker (commented):
Looks good! I must have missed this when migrating to the httpclientprovider from the sdk. Question, what happens is a QueryTimeout is not actively specified in the config editor? The httpClientOptions has a default timeout of 30 seconds (see [here](https://github.com/grafana/grafana-plugin-sdk-go/blob/6d0be6b7893a2bf6297021194a917ad2b52160e9/backend/httpclient/options.go#L83)), so possible safer to guard againt empty string. 
```go
if datasourceSettings.QueryTimeout != "" {
   httpClientOptions.Timeouts.Timeout = datasourceSettings.QueryTimeout
}
```

joshhunt:
@sunker a default value for QueryTimeout is already being set https://github.com/grafana/azure-data-explorer-datasource/blob/e4a4aae78de823a29d5701fb553b4c1df66f9cbf/pkg/azuredx/models/settings.go/#L49-L55

I did some manual testing and this seems to work fine. Is my understanding here correct?

Issue #257, opened 2021-08-11
chore: migrate to Drone


vickyyyyyyy:
We are no longer using CircleCI for most of our builds, including `grafana`: https://drone.grafana.net/grafana/grafana

Migrate from CircleCi to Drone: https://drone.grafana.net/grafana/azure-data-explorer-datasource/settings

sarahzinger:
I'm going to move this to Needs triage because our main branch is currently failing for reasons that I think are not code-related but are CI-related. Maybe we could try to prioritize picking this one up soon.

PR #265, opened 2021-09-29
Use React annotations support

### LINKED ISSUE #253: Annotations not showing up
Hey,

I'm trying to add query annotation as suggested in the Readme.
The query is working fine, but nothing is showing up on the charts.

Other data source annotation work.

Grafana v8.0.6 (68fe9e3431)

ClientEventLog
| project Timestamp=env_time, Text=version , Tags="tag1,tag2"

iwysiu:
Fixes #253 

Also added migrateAnnotation to change old valid annotation queries to the new format.

### DIFF
--- a/src/migrations/annotation.test.ts
+++ b/src/migrations/annotation.test.ts
@@ -31,7 +31,7 @@ const NEW_ANNOTATION = {
   },
 };
 
-describe('ADX: migrateAnnotation', () => {
+describe('migrateAnnotation', () => {
   it('migrates old annotations', () => {
     const migrated = migrateAnnotation(OLD_ANNOTATION);
     expect(migrated).toEqual(NEW_ANNOTATION);
--- a/src/migrations/annotation.test.ts
+++ b/src/migrations/annotation.test.ts
@@ -1,0 +1,47 @@
+import { defaultQuery, EditorMode } from 'types';
+import { migrateAnnotation } from './annotation';
+
+const OLD_ANNOTATION = {
+  database: 'PerfTest',
+  datasource: 'Azure Data Explorer Datasource',
+  enable: true,
+  iconColor: 'red',
+  name: 'New annotation',
+  query: 'PerfTest\n| where $__timeFilter(_Timestamp_) and _val1_ > 15\n| project _Timestamp_, Text=_flag1_',
+  resultFormat: 'table',
+};
+
+const NEW_ANNOTATION = {
+  database: 'PerfTest',
+  datasource: 'Azure Data Explorer Datasource',
+  enable: true,
+  iconColor: 'red',
+  name: 'New annotation',
+  query: 'PerfTest\n| where $__timeFilter(_Timestamp_) and _val1_ > 15\n| project _Timestamp_, Text=_flag1_',
+  resultFormat: 'table',
+  target: {
+    database: 'PerfTest',
+    expression: defaultQuery.expression,
+    pluginVersion: defaultQuery.pluginVersion,
+    query: 'PerfTest\n| where $__timeFilter(_Timestamp_) and _val1_ > 15\n| project _Timestamp_, Text=_flag1_',
+    querySource: EditorMode.Raw,
+    rawMode: true,
+    refId: 'Anno',
+    resultFormat: 'table',
+  },
+};
+
+describe('ADX: migrateAnnotation', () => {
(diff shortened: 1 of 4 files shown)


src/migrations/annotation.ts
```diff
@@ -0,0 +1,23 @@
+import { KustoQuery, defaultQuery } from 'types';
+
+export const migrateAnnotation = (annotation: any) => {
```
joshhunt, line 3: I'm not enthusiastic about introducing any more untyped `any`s into the code - I think it's really easy to ship code that breaks with it, and it's not a pattern I would like to see encouraged further.

Instead is it possible to use `unknown` or `AnnotationQuery<T>`, or a specific type (with all optional fields) that forces us to make sure we're checking the types of things?

src/migrations/annotation.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    pluginVersion: defaultQuery.pluginVersion,
+    query: 'PerfTest\n| where $__timeFilter(_Timestamp_) and _val1_ > 15\n| project _Timestamp_, Text=_flag1_',
+    querySource: EditorMode.Raw,
+    rawMode: true,
+    refId: 'Anno',
+    resultFormat: 'table',
+  },
+};
+
+describe('ADX: migrateAnnotation', () => {
```
joshhunt, line 34: minor nit: everything in these tests are "ADX", so it probably doesnt need the prefix 😅

src/migrations/annotation.test.ts
```diff
@@ -0,0 +1,47 @@
...
+  it('migrates old annotations', () => {
+    const migrated = migrateAnnotation(OLD_ANNOTATION);
+    expect(migrated).toEqual(NEW_ANNOTATION);
+  });
+
+  it('passes through already migrated queries untouched', () => {
+    const newAnnotation = { ...NEW_ANNOTATION };
+    const migrated = migrateAnnotation(newAnnotation);
+
+    // We use .toBe because we want to assert that the object identity did not change!!!
```
joshhunt, line 44: 😇

joshhunt (commented):
Awesome work. Love to see PRs where we fix bugs by a) deleting code b) adding more tests than lines of code we deleted! 🥰

I'll give this a test to see how it goes later on, but this looks all good from me!

src/migrations/annotation.test.ts
iwysiu: done

src/migrations/annotation.ts
iwysiu: I added a specific type since the `database` field makes it not a AnnotationQuery, and I couldn't figure out what type it actually is

src/migrations/annotation.ts
```diff
@@ -0,0 +1,34 @@
...
+    return annotation;
+  }
+
+  const newQuery: KustoQuery = {
+    ...defaultQuery,
+    ...(annotation.target ?? {}),
+    database: annotation.database,
+    query: annotation.query ?? '',
+    rawMode: true,
+    refId: annotation.target?.refId ?? 'Anno',
```
sarahzinger, line 25: Do you know where does 'Anno' come from? Does refId have to be unique? I'm not very familiar with this yet but I see elsewhere we do stuff like:
```
refId: `adx-${interpolatedQuery}`
```

sarahzinger (commented):
Looks really good! I don't know why our build seems to be failing but I see similar problems in the main branch? When I run yarn build locally it seems fine I wonder if this is some kind of circleci issue and we should just migrate to drone already? I moved this ticket out from the backlog and into needs triage since I think this is something we should try to prioritize probably. Not sure what's involved maybe it's as easy as removing some circle files and adding the drone files https://github.com/grafana/azure-data-explorer-datasource/issues/257

src/migrations/annotation.ts
sarahzinger: seems like a chill solution to me! Fun typescript facts just for future reference you can [extend types](https://www.typescriptlang.org/docs/handbook/2/objects.html#extending-types) if you want, or you could edit the original AnnotationQuery to have an optional database property with a question mark.

joshhunt: IIRC `AnnotationQuery`  is typed at the end with `[key: string]: string` or something like that, meaning it allows any named property.

sarahzinger (approved):

PR #266, opened 2021-10-04
Remove circleci code

### LINKED ISSUE #257: chore: migrate to Drone

sarahzinger:
fixes #257 
Looks like Drone is already set up: https://drone.grafana.net/grafana/azure-data-explorer-datasource but was not showing up here on github I assume because circleci is getting in the way. I removed the webhook (along with some other webhooks I didn't see us using...hopefully I didn't remove too much?).

I also added a github action from another datasource for releases and copy/pasta'd the instructions over as I assume this will work the same.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,75 @@
+name: CI
```
sarahzinger, line 1: I'm not 100% we actually use this github action as CI seemed to run before I added this action? But I see it in other datasources so I sort of assume it's needed!

sarahzinger: hmm yeah this seems to add an additional CI step that seems very similar to the setup we already have in drone, should I just remove maybe?

CONTRIBUTING.md
```diff
@@ -51,21 +51,21 @@ Make sure you have the following dependencies installed first:
...
 1. Update the version number in the `package.json` file.
-2. Update the `CHANGELOG.md` with the changes containted in the release.
+2. Update the `CHANGELOG.md` with the changes contained in the release.
 3. Commit the changes to master and push to GitHub.
-4. Create a release branch locally that follows the convention v(major).(minor).x, leaving the patch as `x`. For example:
+4. Create a tag locally that follows the convention v(major).(minor).x, leaving the patch as `x`. For example:
 
    ```bash
-   git checkout -b v3.3.x
+   git tag v3.3.0
```
vickyyyyyyy, line 59: this line is a bit confusing, saying 'leaving the patch as `x`' but then it....doesn't.
Would it make more sense for it to be `git tag v3.3.x`?

PR #274, opened 2021-11-18
On-Behalf-Of Token Authorization

### LINKED ISSUE #107: Feature Request - Authenticate to ADX using user credentials instead of a single application credentials
We'd prefer if the ADX plugin would not use a single App ID for authenticating to ADX, but instead will propagate user credentials for the query.

For us this will enable a few more scenarios that we now restrict due to  this single app authentication.

It'll also make managing permissions easier since users will automatically be restricted to view panels results according to their permissions over the "raw data".

pascaldekloe:
Propagate Grafana client credentials to ADX with a token exchange. The `onBehalfOf` toggle makes the service account impersonate the user by augmenting the access token. See the [developer documentation](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-on-behalf-of-flow) for detail on the concept.

The implementation requires ID tokens <https://github.com/grafana/grafana/pull/42938>.

Resolves #107


andresmgot:
Thank you for the PR @pascaldekloe!

We are going to hold from reviewing this until the dependent PR in Grafana gets resolved since it may affect the solution proposed here.

Regards,

pkg/azuredx/azureauth/auth.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"errors"
+	"strings"
+
+	"github.com/grafana/grafana-plugin-sdk-go/backend"
+)
+
+// BearerToken extracts the client's token.
+func BearerToken(req *backend.QueryDataRequest) (accessToken string, err error) {
+	s, ok := req.Headers["Authorization"]
```
alexanderzobnin, line 13: With our [proposal](https://github.com/grafana/grafana/pull/42422) it can be like
```suggestion
	s, ok := req.Headers["X-ID-Token"]
```

pkg/azuredx/azureauth/auth.go
```diff
@@ -0,0 +1,24 @@
...
+	if !ok {
+		return "", errors.New("no HTTP Authorization value")
+	}
+
+	const prefix = "Bearer "
```
alexanderzobnin, line 18: When using `X-ID-Token`, then we can just return it.
```suggestion
	return s
```

pkg/azuredx/azureauth/auth.go
```diff
@@ -0,0 +1,133 @@
...
+	"strings"
+	"time"
+
+	"github.com/grafana/grafana-plugin-sdk-go/backend"
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/grafana/azure-data-explorer-datasource/pkg/azuredx/models"
+)
+
+var onBehalfOfLatencySeconds = prometheus.NewHistogramVec(prometheus.HistogramOpts{
```
pascaldekloe, line 21: Any idea why this metric won't show up at the Grafana `/metrics` collection @alexanderzobnin?

pascaldekloe:
It is time for a review now @andresmgot. The required changes are in the latest Grafana already. 🥳

pkg/azuredx/azureauth/auth.go
```diff
@@ -0,0 +1,133 @@
...
+
+// QueryDataAuthorization returns an HTTP Authorization-header value which
+// represents the respective request.
+// ⚠️ A zero string must cause appliance of the service account instead.
+//
+// TODO(pascaldekloe): Merge the service principal appliance in here. The entire
+//	authorization decission must be centralised for security reasons alone.
+//	A fully progmatic approach also bypasses the SDK's incorrect use of
+//	http.RoundTripper.
+func (c *ServiceCredentials) QueryDataAuthorization(req *backend.QueryDataRequest) (string, error) {
```
andresmgot, line 48: `QueryData...` implies that this function does the query, I would rename it to `GetOnBehalfOfBearerToken`. Also, I would set the settings and http client as parameters so you don't need the struct (see my comment below in datasource.go:66). Something like: `GetOnBehalfOfBearerToken(settings, req, http.Client)`.

pkg/azuredx/azureauth/auth.go
```diff
@@ -0,0 +1,133 @@
...
+		return "", fmt.Errorf("on-behalf-of grant POST <%q> response: %w", tokenURL, err)
+	}
+	onBehalfOfLatencySeconds.WithLabelValues(strconv.Itoa(resp.StatusCode)).Observe(float64(time.Since(reqStart)) / float64(time.Second))
+	if resp.StatusCode/100 != 2 {
+		var deny struct {
+			Desc string `json:"error_description"`
+		}
+		_ = json.Unmarshal(body, &deny)
+		return "", fmt.Errorf("on-behalf-of grant POST <%q> status %q: %q", tokenURL, resp.Status, deny.Desc)
+	}
```
andresmgot, line 120: I would remove this metric from this PR

pkg/azuredx/datasource.go
```diff
@@ -62,6 +63,11 @@ func NewDatasource(settings backend.DataSourceInstanceSettings) (instancemgmt.In
 	}
 	adx.client = client.New(httpClient)
 
+	adx.serviceCredentials = azureauth.ServiceCredentials{
+		DatasourceSettings: *adx.settings,
+		PostForm:           httpClient.PostForm,
+	}
```
andresmgot, line 69: I would avoid this inner `serviceCredentials` struct for two reasons: the settings are duplicated (`adx.settings` and `adx.serviceCredentials.DatasourceSettings`) and it's only necessary for the on-behalf-of scenario. I would add the settings and the httpClient as parameters for the `GetOnBehalfOfBearerToken` function instead.

pkg/azuredx/datasource.go
```diff
@@ -78,8 +84,13 @@ func (adx *AzureDataExplorer) QueryData(ctx context.Context, req *backend.QueryD
 	backend.Logger.Debug("Query", "datasource", req.PluginContext.DataSourceInstanceSettings.Name)
 	res := backend.NewQueryDataResponse()
 
+	authorization, err := adx.serviceCredentials.QueryDataAuthorization(req)
```
andresmgot, line 87: for readability, I would check the onBehalfOf property here:

```
authHeader := ""
if adx.settings.EnableOnBehalfOf {
  authHeader = adx.GetOnBehalfOfBearerToken(...
}

File: src/components/ConfigEditor/ConnectionConfig.tsx [hunk 8]
```diff
@@ -29,7 +29,10 @@ const ConnectionConfig: React.FC<ConnectionConfigProps> = ({
     if (!jsonData.azureCloud) {
       updateJsonData('azureCloud', AzureCloudType.AzurePublic);
     }
-  }, [jsonData.azureCloud, updateJsonData]);
+    if (!jsonData.oauthPassThru) {
+      updateJsonData('oauthPassThru', true);
```
andresmgot, line 33: I don't think we should enable `oauthPassThru` by default. This is only necessary when enabling the on-behalf-of process.

src/components/ConfigEditor/ConnectionConfig.tsx
```diff
@@ -143,6 +146,30 @@ const ConnectionConfig: React.FC<ConnectionConfigProps> = ({
...
+          <>
+            Propagate Grafana client credentials to ADX with a token exchange. When enabled the service account (Client
+            ID) impersonates the user by augmenting the access token. See the{' '}
+            <a
+              target="_blank"
+              href="https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-on-behalf-of-flow"
+            >
+              developer documentation
+            </a>{' '}
+            for detail on the concept.
```
andresmgot, line 163: I would add a note that OAuth should be enabled and that it requires Grafana > 8.3.3

andresmgot (commented):
Thanks @pascaldekloe! I have done a first review but it would be convenient that someone with more context about this development reviews it. @alexanderzobnin @xlson @sunker can we get another pair of eyes here? (Note that reviews may be delayed due to vacations :))

andresmgot:
btw, you need to merge master here in order to get the CI working.

src/components/ConfigEditor/ConnectionConfig.tsx
pascaldekloe: Good point.

At the moment we do not have a user-interface for this feature. Shall I add another checkbox?

src/components/ConfigEditor/ConnectionConfig.tsx
pascaldekloe: Right. I wanted to verify the configuration and version in the plugin user-interface but got stuck while trying. Is the main configuration available in the plugin somehow?

pkg/azuredx/azureauth/auth.go
```diff
@@ -0,0 +1,133 @@
...
+// QueryDataAuthorization returns an HTTP Authorization-header value which
+// represents the respective request.
+// ⚠️ A zero string must cause appliance of the service account instead.
+//
+// TODO(pascaldekloe): Merge the service principal appliance in here. The entire
+//	authorization decission must be centralised for security reasons alone.
+//	A fully progmatic approach also bypasses the SDK's incorrect use of
+//	http.RoundTripper.
+func (c *ServiceCredentials) QueryDataAuthorization(req *backend.QueryDataRequest) (string, error) {
+	if !c.OnBehalfOf {
```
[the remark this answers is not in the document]
pascaldekloe: It is best to describe what you do in APIs—now how.

pkg/azuredx/azureauth/auth.go
```diff
@@ -0,0 +1,133 @@
...
+		return "", nil
+	}
+
+	user := req.PluginContext.User // do nil-check once for all
+	if user == nil {
+		backend.Logger.Debug("using service principal for non-user request despite on-behalf-of configuration")
```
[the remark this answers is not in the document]
pascaldekloe: We should use the service principal if the request does not come from a user. If for no other reason: what other choice do we have?

> 	// User is the Grafana user making the request.
>	//
>	// Will not be provided if Grafana backend initiated the request,
>	// for example when request is coming from Grafana Alerting.
>	User *User

pkg/azuredx/azureauth/auth.go
```diff
@@ -0,0 +1,133 @@
...
+		return "", nil
+	}
+
+	// Azure requires an ID token (instead of an access token) for the exchange.
+	userToken, ok := req.Headers["X-ID-Token"]
+	if !ok {
+		if _, ok := req.Headers["Authorization"]; !ok {
+			// Grafana permits a mixture of OAuth2 and system accounts.
+			backend.Logger.Warn("user credentials unavailable for data request; continue with service principal despite on-behalf-of configuration", "user", user.Login)
```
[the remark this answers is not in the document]
pascaldekloe: Did you understand the comment above @andresmgot? It is easy to have the plugin fail on system accounts when the on-behalf-of switch is enabled. Some customers may manage datasources and panels with system accounts, and use OAuth2 accounts for the actual day-to-day monitoring. They would be stuck if we treat this like an error. Your call…

pkg/azuredx/azureauth/auth.go
pascaldekloe: QueryDataAuthorization gets you the Authorization of QueryData, just like IceCreamCount gets you the Count of IceCream (and not the IceCream itself).

Again, function names should describe what they do—now how. The actual methods like on-behalf-of and/or bearer-tokens are likely to fluctuate over time.

You *never* need a struct, as everything can be done by passing arguments to functions. The struct here encapsulates a responsibility with easy to understand boundaries. The method call clearly expresses that it provides the authorization for a request.

pkg/azuredx/azureauth/auth.go
pascaldekloe: Metrics are an integral part of the functionality just like logging is. We can take it out of course if you really prefer an extra step.

pkg/azuredx/datasource.go
pascaldekloe: We need the settings for the service-principal and user-forwarding  options too. I have a follow up patch which completes this architecture by moving the `tokenprovider` into ServiceCredentials too. Then authorization is nicely handled in one progmatic function, rather than hidden in layers form another code base. See the TODO for details.

pkg/azuredx/datasource.go
pascaldekloe: It's all about encapsulation. Why pollute the query method with authorization logic?

pkg/azuredx/azureauth/auth.go
pascaldekloe: Started a bug report <https://github.com/grafana/grafana-plugin-sdk-go/issues/439> now.

pkg/azuredx/azureauth/auth.go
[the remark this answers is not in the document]
andresmgot: Okay, yes as I say I am not too familiar with the Azure authentication model. Would this mean that a user could strip their user information from a query and impersonate the service principal?

If that's a yes and it could be a security issue it's probably better for the admin to create two datasources, one for users and another one for service accounts (and fail here). If not, this is fine. Thoughts?

pkg/azuredx/azureauth/auth.go
andresmgot: I know but this is not working atm and if we want to add metrics it would be nice to follow some more metrics, not just this one. So yes please take it out for the moment, we can create another issue to include metrics.

pkg/azuredx/datasource.go
andresmgot: not feeling strongly about this one either

pkg/azuredx/datasource.go
andresmgot: If I just read `authorization, err := adx.serviceCredentials.QueryDataAuthorization(req)` that would mean for me that we are somehow managing all authorization in that method, when really what it's doing is retrieving an optional header (falling back to the default auth method if not configured).

In any case, just naming again.

src/components/ConfigEditor/ConnectionConfig.tsx
andresmgot: I don't think another switch is needed (since you want either both or neither of them AFAIK). Just enable this setting when onBehalfOf is enabled.

src/components/ConfigEditor/ConnectionConfig.tsx
andresmgot: It's possible to retrieve that configuration and hide this switch but probably is not worthy. The note explaining it should be enough.

pkg/azuredx/azureauth/auth.go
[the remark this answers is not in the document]
pascaldekloe: Correct @andresmgot. Since the test query (from the "Save & test" button on the configuration panel) also uses the service principal, regardless of the on-behalf-of checkbox state, we should be fine in terms of predictability.

Just to be clear, we do require the ID token to be present for all users whom login with OAuth2. It is just the service accounts—which must be explicitly created—that bypass the user propagation.

pkg/azuredx/azureauth/auth.go
pascaldekloe: That makes sense. No untested stuff in the commit.

pkg/azuredx/datasource.go
pascaldekloe: You are right. I intend to do "manage all authorization in that method" in a follow up commit, waiting for this to happen first. I can push the PR already for you to see.

src/components/ConfigEditor/ConnectionConfig.tsx
pascaldekloe: Now, the only way to enable user-credential forwarding is by configuring the plugin with the undocumented `oauthPassThru` boolean. To keep that method functional I would have to store the configured value there to recover it's value when on-behalf-of is disabled. Otherwise, the switch would be disabled if the users clicks the on-behalf-of checkbox on and then off again.

Of course I am happy to do so if that is what you want. I recommend creating some sort of UI switch now. It would require less code (in this patch) plus you get to signal the user about conflicting settings more clearly.

I'm awaiting your final decision here.

src/components/ConfigEditor/ConnectionConfig.tsx
pascaldekloe: What do you mean by not worthy? My time is free to you. 😉 ASML gladly pays to have this plugin inform the user, rather than shift responsibility with an increasing list of warnings and instructions.

sunker (commented):
Great to see the progress on this @pascaldekloe. 

Here are some thoughts from me. 
* The OBO flow introduces a new path for handling auth. There is already a package called [tokenprovider](https://github.com/grafana/azure-data-explorer-datasource/blob/0868ee2e2e7f9e543418a4ae73376d885a2af4e9/pkg/azuredx/tokenprovider/token_provider.go). I don't see any reason for not integrating the OBO flow logic into this package so that the OBO token is added by the [auth middleware](https://github.com/grafana/azure-data-explorer-datasource/blob/0868ee2e2e7f9e543418a4ae73376d885a2af4e9/pkg/azuredx/client/auth_middleware.go). You just need to find a way of injecting the id token into the tokenprovider in the query data handler. 
* Currently, an api call to exchange the user token for an OBO token is done every time query data is called. Is that necessary? Feel like this should **only** be necessary if the OBO token has expired or if the X-ID-Token has changed. The tokenprovider package has a cache that can be used once the OBO flow is integrated with that package. 
* I think it needs to be clear what kind of token is being used at all time. If the user has activated OBO in the config page, we should never allow fallback to the system user. So I agree with @andresmgot - if user is not logged in, we should return an error. Consequently, if one wants to create alert queries for example, they should create a new ADX datasource instance without OBO activated.

sunker:
I also wonder, isn't there a method for exchanging the token in the azure sdk for go?

pascaldekloe:
Hi @sunker. It was you whom suggested that the authorization header should be set in the handleQuery method. I agree. What made you change your mind here?

As said before, the auth middleware solution is broken. It violates the http.RoundTripper contract. Changing the request in the client may cause unexpected behaviour, including security issues in the worst case scenario. Again, I have a followup patch which moves the middleware solution into the new authorization method. Did you see the TODO in this patch?

> // TODO(pascaldekloe): Merge the service principal appliance in here. The entire
> //	authorization decission must be centralised for security reasons alone.
> //	A fully progmatic approach also bypasses the SDK's incorrect use of
> //	http.RoundTripper.


Caching the token exchange is super easy. I got it working with [Ristretto](https://pkg.go.dev/github.com/dgraph-io/ristretto#Cache.SetWithTTL) in no time. ASML is doing a performance test now. It's a bit unfortunate that the SDK metrics are broken at the moment. We will share the results of course.

Microsoft released a new Go library with on-behalf-of support a few weeks ago. We could use it instead of the `onBehalfOf` method. Pulling in a dependency with multiple transient dependencies for just a single HTTP POST seems undesirable to me. Especially since we do have a home made TTL cache in the `tokenprovider` package. That cache implementation has race issues b.t.w.. I think we should eliminate it a.s.a.p.. Would Ristretto be OK for you?

sunker:
>It was you whom suggested that the authorization header should be set in the handleQuery method. I agree. What made you change your mind here?

Did I? Possibly I did for the POC, but I don't think that's the path that we should be taking in the final design. 

>As said before, the auth middleware solution is broken. It violates the http.RoundTripper contract. Changing the request in the client may cause unexpected behaviour, including security issues in the worst case scenario. Again, I have a followup patch which moves the middleware solution into the new authorization method. Did you see the TODO in this patch?

I saw it, but I don't understand what you mean. In what way is it violating the http.RoundTripper contract? 

>Microsoft released a new Go library with on-behalf-of support a few weeks ago. We could use it instead of the onBehalfOf method. Pulling in a dependency with multiple transient dependencies for just a single HTTP POST seems undesirable to me.

I was looking for the library but couldn't find it. Can you please share me the link? In the token provider package, we use azcore package and azidentity package from the azure go sdk. I believe the go compiler only links stuff we actually use, so these to packages are only about 1 mb. I assume it's the same with the OBO package. I think I'd prefer to use the package provided by Microsoft instead of doing this with http. 

>That cache implementation has race issues b.t.w.. I think we should eliminate it a.s.a.p.. 

Can you elaborate on this? Please let us know so that we can fix the problems. 

Probably not relevant for you in this PR, but just an FYI: The token provider package in ADX is basically a copy of the token [provider package](https://github.com/grafana/azure-data-explorer-datasource/tree/master/pkg/azuredx/tokenprovider) used by Azure Monitor in core Grafana. I believe the only difference is that the package used by Azure Monitor has support for Managed Identity auth. We plan to move this package into a separate repo so that it can be shared by all Azure plugins sometime soon. Doing so removes duplication if code, but it also enables Managed Identity support in ADX. It should also make it possible to use OBO in for example Azure Monitor. That's yet another reason to move the OBO logic into the tokenprovider package.

pascaldekloe:
>	// RoundTrip should not modify the request, except for
>	// consuming and closing the Request's Body. RoundTrip may
>	// read fields of the request in a separate goroutine. Callers
>	// should not mutate or reuse the request until the Response's
>	// Body has been closed.
> — https://pkg.go.dev/net/http#RoundTripper

The Microsoft [library](https://github.com/AzureAD/microsoft-authentication-library-for-go) is not part of the SDK yet. It is still in early stages of development. Adding this code will "enrich" Grafana with the following dependencies.

*	github.com/golang-jwt/jwt v3.2.1+incompatible
*	github.com/google/uuid v1.1.1
*	github.com/kylelemons/godebug v1.1.0
*	github.com/montanaflynn/stats v0.6.6 // indirect
*	github.com/pkg/browser v0.0.0-20210115035449-ce105d075bb4
*	golang.org/x/sys v0.0.0-20210124154548-22da62e12c0c // indirect

The cache races are due to a mix of atomic and mutex use, i.e., in Go the mutex do not lock atomic operation and verse visa. No need to fix it. Go is fortunate with lots of excellent libraries for TTL caching.

Ah, I see now why you are hesitant about my attempt to unify and simplify the authorization. On the other hand, if I hand over a package with both the on-behalf-of and the token provider then that should be exactly what you want in such shared library right? Even better; this gives you a way out of this middleware setup.

sunker:
Okay thanks, I'll discuss OBO lib vs http with the team and get back to you. 

> The cache races are due to a mix of atomic and mutex use, i.e., in Go the mutex do not lock atomic operation and verse visa. 

@kostrse you've worked with the cache. Is this something we'd like to look into? 

> Ah, I see now why you are hesitant about my attempt to unify and simplify the authorization. On the other hand, if I hand over a package with both the on-behalf-of and the token provider then that should be exactly what you want in such shared library right? Even better; this gives you a way out of this middleware setup.

@pascaldekloe the OBO of logic in this PR does not unify auth. It adds yet another way of handling auth which makes the code much harder to follow. Can you explain in what way the auth middleware is violating the htth.RoundTripper contract?

src/components/ConfigEditor/ConnectionConfig.tsx
pascaldekloe: 🛎 @andresmgot @sunker

I think the UI should provide a "User Credentials" select with the options: "None" [service principal], Forward [oauthPassTru] and "On Behalf Of".

Since the `oauthPassThru` option is undocumented, I suggest to drop support a.s.a.p. in favour of a new field, corresponding to the select options. E.g. `"user-creds": "on-behalf-of"`.

pkg/azuredx/datasource.go
pascaldekloe: Since @sunker also felt on this issue I added the followup in here (to see and discuss).

andresmgot:
Thanks @pascaldekloe. @sunker is currently OOO so he will provide more feedback once he is back. In the meantime, can you add docs about the feature? In the main README should be fine.

Also, the original conversations mentioned providing a testing environment so we could test the feature. Is that something you can still provide? Thanks.

andresmgot:
BTW, you still need to merge master here to get the CI working.

pascaldekloe:
> BTW, you still need to merge master here to get the CI working.

I merged twice now. 🤨 The last one—commit 8caa2c7—sits right above your two messages.

>  In the meantime, can you add docs about the feature? In the main README should be fine.

Sure. Users should be able to find its existence with search engines too. 👍

> Also, the original conversations mentioned providing a testing environment so we could test the feature. Is that something you can still provide?

I offered access to speed things up months ago. I also recommended you to create a test environment with company [Grafana] accounts. You need one for testing and development.

andresmgot:
> I merged twice now. raised_eyebrow The last one—commit [8caa2c7](https://github.com/grafana/azure-data-explorer-datasource/commit/8caa2c7b31348826f2051f215f5cb49d593c47a3)—sits right above your two messages.
> 

Ah, yes, there is a lint issue then that you need to fix. Can you access https://drone.grafana.net/grafana/azure-data-explorer-datasource/104/1/4? The error is:

```
  /drone/src/src/components/ConfigEditor/ConnectionConfig.tsx
    160:13  error  Using target="_blank" without rel="noreferrer" is a security risk: see https://html.spec.whatwg.org/multipage/links.html#link-type-noopener  react/jsx-no-target-blank
  
  ✖ 1 problem (1 error, 0 warnings)
```

> I offered access to speed things up months ago. I also recommended you to create a test environment with company [Grafana] accounts. You need one for testing and development.

Can you point us to the email with that info? (or send it again). Yes, we will probably need to test with our accounts, once there is documentation we can try to follow it to also check that. Thanks.

src/components/ConfigEditor/ConnectionConfig.tsx
andresmgot: I don't think we need to add support to only oauthPassThru (since as you say, it's undocumented so no one should be using that atm). The switch below should enable/disable both, `oauthPassThru` and `onBehalfOf` we can add support for only oauthPassThru later if requested.

src/components/ConfigEditor/ConnectionConfig.tsx
andresmgot: I mean maintaining (not only implementing) the code that retrieves the configuration, checks and compares the version and enables/disables this input. The notice with the version should be enough.

pascaldekloe:
The recommendation for an Azure test account came across in our very first meeting, somewhere around August I believe.
You verifying the instructions by actually making an account is a good idea. 🙂 You need a credit card to do so. Charges apply. Where should I write the instructions?

src/components/ConfigEditor/ConnectionConfig.tsx
```diff
@@ -29,7 +29,10 @@ const ConnectionConfig: React.FC<ConnectionConfigProps> = ({
     if (!jsonData.azureCloud) {
       updateJsonData('azureCloud', AzureCloudType.AzurePublic);
     }
-  }, [jsonData.azureCloud, updateJsonData]);
+    if (!!jsonData.oauthPassThru !== !!jsonData.onBehalfOf) {
+      updateJsonData('oauthPassThru', !!jsonData.onBehalfOf);
+    }
+  }, [jsonData.azureCloud, jsonData.oauthPassThru, jsonData.onBehalfOf, updateJsonData]);
```
andresmgot, line 35: I think it's simpler to just set `oauthPassThru` in the function of the line 174, isn't it?

pascaldekloe: UseEffect is also called on the initial value, which applies the parameter binding to the (provisioning) configuration as well. I know little about React so feel free to correct me there.

sunker: If you provision ADX, this will only work if you actually go in and open the configuration page, and that is not ideal. Instead of setting the `oauthPassThru` flag on mount of this page, it's better to set this when changing the `onBehalfOf` switch (as Andrés was suggesting). You can also add a new section in the readme with examples of how to provision ADX with and without OBO. 

Example without: 
```yaml
  - name: Azure Data Explorer
    type: grafana-azure-data-explorer-datasource
    access: proxy
    basicAuth: false
    editable: true
    jsonData:
      clientId: '<your client id>'
      clusterUrl: '<your cluster url>'
      tenantId: '<your tenant id>'
      tlsAuth: false
      tlsAuthWithCACert: false
      defaultDatabase: '<your deault database>'
    secureJsonData:
      clientSecret: '<your client secret>'
      tlsCACert: ''
      tlsClientCert: ''
      tlsClientKey: ''
```

With OBO enabled: 
```yaml
  - name: Azure Data Explorer
    type: grafana-azure-data-explorer-datasource
    access: proxy
    basicAuth: false
    editable: true
    jsonData:
      oauthPassThru: true
      onBehalfOf: true
      clientId: '<your client id>'
      clusterUrl: '<your cluster url>'
      tenantId: '<your tenant id>'
      tlsAuth: false
      tlsAuthWithCACert: false
      defaultDatabase: '<your deault database>'
    secureJsonData:
      clientSecret: '<your client secret>'
      tlsCACert: ''
      tlsClientCert: ''
      tlsClientKey: ''
```

See [cloudwatch docs](https://github.com/grafana/grafana/blob/main/docs/sources/datasources/aws-cloudwatch/provision-cloudwatch.md#L18) for inspiration

src/components/ConfigEditor/ConnectionConfig.tsx
```diff
@@ -146,6 +149,31 @@ const ConnectionConfig: React.FC<ConnectionConfigProps> = ({
...
         tooltip={clientSecretTooltip}
       />
+
+      <InlineField
+        label="Use On-Behalf-Of"
+        labelWidth={26}
+        tooltip={
+          <>
+            Propagate Grafana client credentials to ADX with a token exchange. When enabled the service account (Client
+            ID) impersonates the user by augmenting the access token. See the{' '}
```
sunker, line 159: I like this, but the part about grafana version should be in the documentation and not in a tooltip.

doc/on-behalf-of.md
```diff
@@ -0,0 +1,7 @@
+# On-Behalf-Of Authorization
+
+Configure Grafana to use OAuth2 with Azure Active Directory as [documented](https://grafana.com/docs/grafana/latest/auth/azuread/). The `[auth.azuread]` `scopes` (or `$GF_AUTH_AZUREAD_SCOPES`) must contain “openid email profile”.
+
+ID tokens must be enabled with a checkbox found on the [Azure portal](https://portal.azure.com/) under “App Registrations” → the respective application → “Manage” → “Authentication”.
+
+In addition to the “Microsoft Graph” `User.Read`, a special “Azure Data Explorer” `user_consent` permission must be enabled on the [Azure portal](https://portal.azure.com/)  under “App Registrations” → the respective application → “Manage” → “API permissions”.
```
sunker, line 7: I think we also need to say explicitly that it's not recommended to create alert rules for an ADX data source instance that is configured to use OBO.

pkg/azuredx/azureauth/auth.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+// QueryDataAuthorization returns an HTTP Authorization-header value which
+// represents the respective request.
+func (c *ServiceCredentials) QueryDataAuthorization(req *backend.QueryDataRequest) (string, error) {
+	switch {
+	case c.OnBehalfOf:
+		return c.queryDataOnBehalfOf(req)
+
+	// TODO(pascaldekloe): case c.ClientCreds [oauthPassThru]
```
sunker, line 60: Remove this?

pkg/azuredx/azureauth/auth.go
```diff
@@ -0,0 +1,152 @@
...
+
+	default:
+		backend.Logger.Debug("using service principal token for data request")
+		ctx := context.Background()
```
sunker, line 64: QueryDataAuthorization should accept context as an argument and use that instead.

pkg/azuredx/azureauth/auth.go
```diff
@@ -0,0 +1,152 @@
...
+	params := make(url.Values)
+	params.Set("client_id", c.ClientID)
+	params.Set("client_secret", c.Secret)
+	params.Set("grant_type", "urn:ietf:params:oauth:grant-type:jwt-bearer")
+	params.Set("assertion", userToken)
+	params.Set("scope", c.ClusterURL+"/.default")
+	params.Set("requested_token_use", "on_behalf_of")
+
+	// https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-v2-protocols#endpoints
+	tokenURL := "https://login.microsoftonline.com/" + url.PathEscape(c.TenantID) + "/oauth2/v2.0/token"
```
sunker, line 115: ADX supports multiple clouds, but it looks like this url is targeting Azure public cloud only. Do you know if this api is available in all clouds? If not, we might have to disable OBO for all national clouds. If it is available in all clouds, you need to resolve the host based on the clouds. See [here](https://github.com/grafana/azure-data-explorer-datasource/blob/master/pkg/azuredx/tokenprovider/token_provider.go#L56-L68) how it's done in the token provider.

src/components/ConfigEditor/ConnectionConfig.tsx
pascaldekloe: I believe that is exactly what happens now @sunker. No need to specify `oauthPassThru`—it will follow whatever the `onBehalfOf` is.

pkg/azuredx/datasource.go
```diff
@@ -78,16 +76,26 @@ func (adx *AzureDataExplorer) QueryData(ctx context.Context, req *backend.QueryD
...
+		res.Responses[q.RefID] = adx.handleQuery(q, req.PluginContext.User, authorization)
 	}
 
 	return res, nil
 }
 
 // CheckHealth handles health checks
 func (adx *AzureDataExplorer) CheckHealth(ctx context.Context, req *backend.CheckHealthRequest) (*backend.CheckHealthResult, error) {
-	err := adx.client.TestRequest(adx.settings, models.NewConnectionProperties(adx.settings, nil))
+	authorization, err := adx.serviceCredentials.ServicePrincipalAuthorization(ctx)
```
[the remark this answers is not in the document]
pascaldekloe: CheckHealthRequest does not provide a user. You are right though. This API needs fixing. Specifically: `CheckHealth` should call `QueryData`. Not something part of this PR though. All I do is make the behaviour explicit.

src/components/ConfigEditor/ConnectionConfig.tsx
pascaldekloe: Can you be more specific on where things should be?

pkg/azuredx/azureauth/auth.go
pascaldekloe: I am not sure whether this mess applies here too. The documentation form the URL in the comment above is quite clear on it's location. I will ask a contact at Microsoft to be sure…

pascaldekloe: It did indeed.

pkg/azuredx/datasource.go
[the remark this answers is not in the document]
sunker: Will you take a stab at adding that to the API? If not, please create an issue in the [plugin-sdk](https://github.com/grafana/grafana-plugin-sdk-go)

src/components/ConfigEditor/ConnectionConfig.tsx
sunker: You can add something like `Note: # On-Behalf-Of Authorization is only compatible with Grafana 8.3.4 or later.` in the On-Behalf-Of Authorization section in the documentation.

pkg/azuredx/azureauth/auth.go
sunker: Not sure I follow - it this api available in all clouds?

src/components/ConfigEditor/ConnectionConfig.tsx
sunker: Yeah but only if you actually open the configuration page? That's when the effect is triggered. When you provision the app, you don't want to go into the config page - you want to be able to go straight to the dashboard and have a working config.

pascaldekloe: I believe provisioning works without opening the configuration page now. Please let me know if it doesn't.

pkg/azuredx/datasource.go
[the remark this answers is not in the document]
pascaldekloe: Sure.

src/components/ConfigEditor/ConnectionConfig.tsx
pascaldekloe: What is "the documentation"? Can you be specific?

src/components/ConfigEditor/ConnectionConfig.tsx
pascaldekloe: OK, a duplicate provisioning file prevented me from seeing this. Yeah, one way would be to document the workaround.
Does the React setup really prevent us from fixing the issue? Where does the plugin receive the configuration?

sunker: [This](https://github.com/grafana/azure-data-explorer-datasource/blob/91fd99784426b32e490ba8bb1dec3fc5e3808947/pkg/azuredx/datasource.go#L33) is where the plugin receives new settings. In this method you could check if user has bad conigurations, i.e if `onBehalfOf` whereas `oauthPassThru` is missing (or set to false). In that case, you can log an error. 

Please add a new section with the header `Configure the data source with provisioning` at the bottom of the readme file. Add the two examples from above. You can use the documentation from [timestream data source](https://github.com/grafana/timestream-datasource/blob/main/src/README.md#configure-the-data-source-with-provisioning) as inspiration when formatting the markdown.

src/components/ConfigEditor/ConnectionConfig.tsx
sunker: In doc/on-behalf-of.md. Add `⚠️ Only compatible with Grafana 8.3.4 or later` above `⚠️ Grafana alert-rules do not work with on-behalf-of authorization.`. It will look a bit messy with two warnings next to each other, but we can improve documentation later.

pkg/azuredx/azureauth/auth.go
sunker: @pascaldekloe

src/components/ConfigEditor/ConnectionConfig.tsx
pascaldekloe: Setting `oauthPassThru` in the Go code won't affect the configuration unfortunately. The user should not be bothered by such internal detail. There must be a place in this JavaScript mess where the configuration is read and applied. Is it still reachable, or are we lost in layers of abstraction?

sunker: >Setting oauthPassThru in the Go code won't affect the configuration unfortunately.

I know it won't. We can't prevent user's from adding bad configurations while using provisioning, but we can print a log message to let them know that the config was bad. This is mainly for the provisioning story. 

If the user configures the data source through the UI (not provisioning), we can prevent them from having bad configuration using code. You should make sure to sync the `oauthPassThru` with the `onBehalfOf` prop on line 188 in the ConnectionConfig.tsx.

pkg/azuredx/azureauth/auth.go
pascaldekloe: Hard to say for sure. Needs testing. This is a new feature, so we won't break anything. The worst case is that it does not work. I say collect the knowledge with bug reports in the spirit of open-source development.

doc/on-behalf-of.md
```diff
@@ -0,0 +1,50 @@
+# On-Behalf-Of Authorization
+
+The feature can be enabled with the checkbox from the plugin configuration screen.
+⚠️ When provisioning, you also need to set `oauthPassThru` to `true`, such as the following example has done.
```
sunker, line 4: I feel like this section has too many exclamation marks. Can you remove this one?

doc/on-behalf-of.md
```diff
@@ -0,0 +1,50 @@
...
+      defaultDatabase: <your default database>
+    secureJsonData:
+      clientSecret: <your client secret>
+      tlsCACert: ''
+      tlsClientCert: ''
+      tlsClientKey: ''
+    version: 1
+```
+
+⚠️ Only compatible with Grafana 8.3.4 or later above.
```
sunker, line 31: Move this sentence to the top of the section (right under the header).

doc/on-behalf-of.md
```diff
@@ -0,0 +1,50 @@
...
+⚠️ Grafana alert-rules do not work with on-behalf-of authorization.
```
sunker, line 32: Replace exclamation mark with: `> **Note:** Grafana alert-rules...`

doc/on-behalf-of.md
```diff
@@ -0,0 +1,50 @@
...
+
+
+## Setup
+
+Configure Grafana to use OAuth2 with Azure Active Directory as [documented](https://grafana.com/docs/grafana/latest/auth/azuread/). The `[auth.azuread]` `scopes` (or `$GF_AUTH_AZUREAD_SCOPES`) must contain “openid email profile”.
+
+ID tokens must be enabled with a checkbox found on the [Azure portal](https://portal.azure.com/) under “App Registrations” → the respective application → “Manage” → “Authentication”.
+
+In addition to the “Microsoft Graph” `User.Read`, a special “Azure Data Explorer” `user_impersonation` permission must be enabled on the [Azure portal](https://portal.azure.com/)  under “App Registrations” → the respective application → “Manage” → “API permissions”. ⚠️ “Admin consent” is required.
+
```
sunker, line 42: Please rewrite this to a _step by step_ setup. E.g:
1. Configure Grafana to use OAuth2...
2. Enable ID tokens.
3. Allow user impersonation permission
4. Enable "Admin consent" under “App Registrations” → the respective application → “Security” → “Permissions”.

Since the admin consent part requires an actual Azure admin to make changes in the account I think you should elaborate a bit on this. You can add something like this _Enabling "Admin consent" will grant consent on behalf of all users in the current tenant, ensuring the users will not be required to consent when using the application._

pkg/azuredx/azureauth/auth.go
```diff
@@ -0,0 +1,196 @@
...
+	params.Set("client_id", c.ClientID)
+	params.Set("client_secret", c.Secret)
+	params.Set("grant_type", "urn:ietf:params:oauth:grant-type:jwt-bearer")
+	params.Set("assertion", userToken)
+	params.Set("scope", c.ClusterURL+"/.default")
+	params.Set("requested_token_use", "on_behalf_of")
+	reqBody := strings.NewReader(params.Encode())
+
+	// https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-v2-protocols#endpoints
+	tokenURL := fmt.Sprintf("https://%s%s/oauth2/v2.0/token", authorityBaseURL(c.AzureCloud), url.PathEscape(c.TenantID))
```
sunker, line 158: The `authorityBaseURL` func already resolves a url prefixed with `https://` so I believe you can remove the hard coded protocol from here.

pkg/azuredx/datasource.go
```diff
@@ -37,57 +37,62 @@ func NewDatasource(settings backend.DataSourceInstanceSettings) (instancemgmt.In
 	if err != nil {
 		return nil, err
 	}
+	// TODO(pascaldekloe): We need a way to set oauthPassThru in the plugin instead.
```
sunker, line 40: still relevant?

sunker (commented):
OBO stopped working for me after latest changes. Presumably because protocols have been added twice OBO token exchange url.

doc/on-behalf-of.md
pascaldekloe: Admin consent is enabled by default already, but OK. Microsoft has plenty of documentation on how to configure the consent. What we need to provide is a checklist (as provided here).

doc/on-behalf-of.md
```diff
@@ -0,0 +1,53 @@
...
+      defaultDatabase: <your default database>
+    secureJsonData:
+      clientSecret: <your client secret>
+      tlsCACert: ''
+      tlsClientCert: ''
+      tlsClientKey: ''
+    version: 1
+```
+
+**note***: Grafana alert-rules do not work with on-behalf-of authorization.
```
sunker, line 33: After closer consideration, this is actually not fully accurate. Please change this to:
_It is not recommended to setup [alerts](https://grafana.com/docs/grafana/latest/alerting/) when the data source is configured to use on-behalf-of authorization. This is because the alert rule will no longer work as soon as the user that created the rule is logged out from Grafana._

doc/on-behalf-of.md
sunker: It should be clear to the user why admin consent is necessary. As said before, add this text:
_Enabling "Admin consent" will grant consent on behalf of all users in the current tenant, ensuring the users will not be required to consent when using the application._

pkg/azuredx/datasource.go
sunker: The plugin will not be able to change json data config at this part of the life cycle, so you can remove the TODO. Even though this is not ideal, it's now clearly stated in the docs how to configure OBO using provisioning.

src/components/ConfigEditor/ConnectionConfig.tsx
```diff
@@ -160,6 +163,31 @@ const ConnectionConfig: React.FC<ConnectionConfigProps> = ({
...
+            </a>{' '}
+            for detail on the concept. This feature requires a Grafana version 8.3.4 or later.
+          </>
+        }
+      >
+        <InlineSwitch
+          id="adx-on-behalf-of"
+          value={jsonData.onBehalfOf}
+          onChange={(ev: React.ChangeEvent<HTMLInputElement>) => updateJsonData('onBehalfOf', ev.target.checked)}
+        />
```
sunker, line 189: Change this to: 

```jsx
<HorizontalGroup>
    <InlineSwitch
            id="adx-on-behalf-of"
            value={jsonData.onBehalfOf}
            onChange={(ev: React.ChangeEvent<HTMLInputElement>) => updateJsonData('onBehalfOf', ev.target.checked)}
          />
   <span>⚠️ This feature is in beta</span>
</HorizontalGroup>
```

doc/on-behalf-of.md
```diff
@@ -0,0 +1,53 @@
+# On-Behalf-Of Authorization
+
+⚠️ Only compatible with Grafana 8.3.4 or later above.
```
sunker, line 3:
```suggestion
**note***: The feature is in beta, which means changes to the API and the user experience might occur.

⚠️ Only compatible with Grafana 8.3.4 or later above.
```

sunker (commented):
Thanks for applying changes. Added a few more comments. Did you hear back from MS regarding national clouds?

sunker:
@grafana/docs-squad can you please review the doc changes?

osg-grafana:
> @grafana/docs-squad can you please review the doc changes?

Added you, @GrafanaWriter, to triage this.

achatterjee-grafana:
I can review the doc component.

pkg/azuredx/azureauth/auth.go
pascaldekloe: We decided that the feature is unsupported.

sunker (commented):
More comments. Now that time is of essence, please make sure you test the code manually before pushing it.

src/components/ConfigEditor/ConnectionConfig.tsx
```diff
@@ -160,6 +163,32 @@ const ConnectionConfig: React.FC<ConnectionConfigProps> = ({
...
+        <span>⚠️ This feature is in beta</span>
```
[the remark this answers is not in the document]
pascaldekloe: OK. Will keep them open. No testing was applied here. Fixed now.

doc/on-behalf-of.md
```diff
@@ -0,0 +1,58 @@
...
+      clientSecret: <your client secret>
+      tlsCACert: ''
+      tlsClientCert: ''
+      tlsClientKey: ''
+    version: 1
+```
+
+**note***: Private clouds (like "chinaazuremonitor" or "govazuremonitor") are unsupported with on-behalf-of authorization. They may, or may not work.
+
+**note***: It is not recommended to setup alerts when the data source is configured to use on-behalf-of authorization. This is because the alert rule will no longer work as soon as the user that created the rule is logged out from Grafana.
```
achatterjee-grafana, line 37:
```suggestion
We do not support on-behalf-of authorization for private clouds (like "chinaazuremonitor" or "govazuremonitor") at this time.

**Note:***: Do not set up alerts when your data source is configured to use the on-behalf-of authorization. Alert rules will not work once the user who created the rule logs off from Grafana. 
```

achatterjee-grafana (commented):
Added some more doc reviews.

README.md
```diff
@@ -158,6 +158,8 @@ If the command succeeds you should get a result like this:
 
 6. Click the `Save & Test` button. After a few seconds once Grafana has successfully connected then choose the default database and save again.
 
+The on-behalf-of switch makes Grafana propagate client credentials to ADX with a token exchange. When enabled the service account (Client ID) impersonates the user by augmenting the access token. See `./doc/on-behalf-of.md` for details.
```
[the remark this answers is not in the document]
pascaldekloe: Relative linking does not work like that in source code repositories like GitHub.

andresmgot (approved):
Thanks for applying all the comments. I gave this a try and worked for me so let's merge and iterate further if needed.

PR #276, opened 2021-11-19
Replace query editor with the one from @grafana/ui

andresmgot:
Loading this custom `KustoCodeEditor` causes a crash if some other monaco editor has been loaded before (and viceversa). As far as I understood that code editor is only used to support `KQL` but since Grafana 8.0 that's supported in the `QueryEditor` from `@grafana/ui`.

If someone with Grafana <8.0 uses the new code editor, they will just lose syntax highlighting. I'd assume that's fine (rather than trying to add some conditional logic to fallback to the previous editor). WDYT? 

Fixes: https://github.com/grafana/support-escalations/issues/1345


sarahzinger (approved):
This seems like a good idea to me!

What would you think about:
- updating the package.json to specify the version of grafana/ui this wants to run on?
- adding a line about this for the changelog
- and then we can release this as a major version release with breaking changes?

andresmgot:
> * updating the package.json to specify the version of grafana/ui this wants to run on?

Sounds good!

> * adding a line about this for the changelog

How can I do that, adding a tag would work in this repo as well?

> * and then we can release this as a major version release with breaking changes?

I wouldn't say this is a breaking change? It doesn't break any query nor require migration steps, it just updates the look and feel. Maybe bumping the minor?

andresmgot:
> > * updating the package.json to specify the version of grafana/ui this wants to run on?
> 
> Sounds good!
> 

Tried that but it didn't work, I got a nasty compilation error :disappointed: I assume it's related to using Node 16. Will do that some other time :upside_down_face:

vickyyyyyyy (approved):
LGTM!

Would be great when we have e2e tests to test that this new editor works as expected

PR #277, opened 2021-11-19
chore: fix bug label for issue template

vickyyyyyyy:
Looks like the syntax for the label isn't right (https://github.com/grafana/grafana/blob/main/.github/ISSUE_TEMPLATE/1-bug_report.md). Not sure how it works for the Grafana repo (maybe they have a separate workflow for the labels?) but looks like this syntax will fix it for our external plugins.

**Before:**
Adding a bug report won't attach a label to the issue.

**After:**
Adding a bug report will attach the labels `datasource/ADX` and `type/bug` to the issue.

Notes: not sure what's up with the whitespace added and the diff with `Others`, looks like it's a linting thing

### DIFF
--- a/.github/ISSUE_TEMPLATE/1-bug_report.md
+++ b/.github/ISSUE_TEMPLATE/1-bug_report.md
@@ -1,7 +1,7 @@
 ---
 name: Bug report
 about: Report a bug you found when using this plugin
-labels: 'type: bug'
+labels: type/bug
 ---
 
 <!--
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/1-bug_report.md
```diff
@@ -1,7 +1,7 @@
 ---
 name: Bug report
 about: Report a bug you found when using this plugin
-labels: 'type: bug'
+labels: datasource/ADX, type/bug
```
andresmgot, line 4: same issue about the array

andresmgot (approved):

PR #299, opened 2021-12-21
Fix macro regex on column with hyphens

### LINKED ISSUE #296: $__timeFilter on a column that has dashes in its name does not work
**What happened**:

I get the following weird error when using query builder in grafana with azure data explorer data source:

```
Failed to execute query: Azure HTTP "400 BadRequest": 
"Request is invalid and cannot be processed: Syntax error: SYN0002: A recognition error occurred. [line:position=2:107]. 
Query: '['fridge-telemetry-from-hub']\n| where TimeGenerated >= datetime(2021-11-20T23:00:00Z) and TimeGenerated <= datetime(2021-11-22T22:59:59Z)(messageProperties_iothub-creation-time-utc)\n| summarize by '"
```

The query builder would work and return results from the underlying adx resource.

See the screenshot below. Likely have an adx table with datetime column that has dashes in it's name.

**Screenshots**

image

I think the issue is in this regexp, the column names that have dashes do not work with the ``$__timeFilter(column-name)`` syntax:

https://github.com/grafana/azure-data-explorer-datasource/blob/919c9dbdeae97589ceb3ef2d70a1d07fd7174c9e/pkg/azuredx/models/macro.go#L36-L38

image

**Environment**:

- Grafana version: Grafana v8.3.3 (30bb7a93ca)
- Plugin version: 3.6.1
- OS Grafana is installed on: linux container
- User OS & Browser: Windows 10, Edge
- Others:

fridgepoet:
Columns with any hyphens were not captured in the macro regex

Fixes #296


andresmgot (commented):
The change looks good to me but let's wait until I merge the other fix so we can rebase/merge here.

andresmgot (approved):
LGTM

pkg/azuredx/models/macro_test.go
```diff
@@ -83,6 +83,17 @@ func TestMacroData_Interpolate(t *testing.T) {
...
+		{
+			name: "should parse $__timeFilter(value-with-hyphens)",
+			macroData: NewMacroData(&backend.TimeRange{
+				From: fromTime,
+				To:   toTime,
+			}, 0),
+			errorIs:   assert.NoError,
+			query:     "$__timeFilter(value-with-hyphens)",
+			returnIs:  assert.Equal,
+			returnVal: fmt.Sprintf("value-with-hyphens >= %v and value-with-hyphens <= %v", fromString, toString),
```
astafan8, line 95: just checking: will the value-with-hyphens be escaped in the resulting kusto query? i found out that with the dashed columns (or table names or variable names) kusto understand only if they are inside `[` and `'`, so like this ``['value-with-hyphens']``.

fridgepoet: Ah no I don't think so. Thanks for pointing that out. 

It's documented [here](https://docs.microsoft.com/en-us/azure/data-explorer/kusto/query/schema-entities/entity-names#identifier-quoting), right?

astafan8: @fridgepoet yes, i think you found the right doc

PR #300, opened 2021-12-22
Fix: Include quotes when identifier has space, dot, or dash

### LINKED ISSUE #296: $__timeFilter on a column that has dashes in its name does not work

fridgepoet:
Need to at least quote identifiers with spaces, dots, or dashes.
https://docs.microsoft.com/en-us/azure/data-explorer/kusto/query/schema-entities/entity-names#identifier-quoting

Fixes #296

### DIFF
--- a/pkg/azuredx/models/macro.go
+++ b/pkg/azuredx/models/macro.go
@@ -35,7 +35,7 @@ func NewMacroData(tr *backend.TimeRange, intervalMS int64) MacroData {
 // macroRE is a regular expression to match available macros
 var macroRE = regexp.MustCompile(`\$__` + // Prefix: $__
 	`(timeFilter|timeFrom|timeTo|timeInterval)` + // one of macro root names
-	`(\([a-zA-Z0-9_-]*?\))?`) // optional () or optional (someArg)
+	`(\([a-zA-Z0-9_ .-]*?\))?`) // optional () or optional (someArg)
 
 // Interpolate replaces macros with their values for the given query.
 func (md MacroData) Interpolate(query string) (string, error) {
@@ -55,7 +55,7 @@ func (md MacroData) Interpolate(query string) (string, error) {
 		}
 		arg := ""
 		if len(varSplit) > 1 {
-			arg = varSplit[1]
+			arg = quoteForSpacesDotsDashes(varSplit[1])
 		}
 		return funcToCall(arg, md)
 	}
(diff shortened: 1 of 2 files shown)


pkg/azuredx/models/macro.go
```diff
@@ -35,7 +35,7 @@ func NewMacroData(tr *backend.TimeRange, intervalMS int64) MacroData {
 // macroRE is a regular expression to match available macros
 var macroRE = regexp.MustCompile(`\$__` + // Prefix: $__
 	`(timeFilter|timeFrom|timeTo|timeInterval)` + // one of macro root names
-	`(\([a-zA-Z0-9_-]*?\))?`) // optional () or optional (someArg)
+	`(\([a-zA-Z0-9_ .-]*?\))?`) // optional () or optional (someArg)
```
andresmgot, line 38: nit: for readability I would use `\s\.`

pkg/azuredx/models/macro.go
```diff
@@ -66,6 +66,13 @@ func (md MacroData) Interpolate(query string) (string, error) {
 	return interpolated, nil
 }
 
+func quoteForSpacesDotsDashes(s string) string {
+	if strings.ContainsAny(s, " .-") {
```
andresmgot, line 70: can you add a link to the docs you found here as a comment? (for future readers)

andresmgot (approved):
Great, thanks!

pkg/azuredx/models/macro.go
fridgepoet: ok! how about with just `\s.-`?

andresmgot (approved):

PR #308, opened 2022-01-25
Add accessibility to connection config

fridgepoet:
Improves accessibility on a couple of fields in the connection configuration view.

### DIFF
--- a/src/components/ConfigEditor/ConnectionConfig.tsx
+++ b/src/components/ConfigEditor/ConnectionConfig.tsx
@@ -70,7 +70,7 @@ const ConnectionConfig: React.FC<ConnectionConfigProps> = ({
         htmlFor="azure-cloud-type"
       >
         <Select
-          id="azure-cloud-type"
+          inputId="azure-cloud-type"
           options={azureClouds}
           value={azureClouds.find((v) => v.value === jsonData.azureCloud)}
           onChange={(change: SelectableValue<AzureCloudType>) =>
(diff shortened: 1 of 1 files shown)


src/components/ConfigEditor/ConnectionConfig.tsx
```diff
@@ -62,6 +62,7 @@ const ConnectionConfig: React.FC<ConnectionConfigProps> = ({
     <FieldSet label="Connection Details">
       <InlineField label="Azure cloud" labelWidth={26} tooltip="Select an Azure Cloud." required>
         <Select
+          aria-label="Azure cloud"
```
andresmgot, line 65: To avoid misusing aria-labels it's generally better to use `data-testid` for the e2e tests. It also useful to use selectors (so the text is saved in a variable) and have the label pointing to that selector. See athena for example: https://github.com/grafana/athena-datasource/blob/main/src/ConfigEditor.tsx#L91

Also, I guess it's probably fine to have this changes with your other PR (a draft for the moment) so you don't have inter-dependent PRs (unless the main PR grows too much),

fridgepoet: These two particular fields weren't actually accessible I believe (according to FastPass).
For the e2e PR, I'm happy to add `data-testid` for these fields, but it will not improve the accessibility. 

I prefer to keep PRs as small and as focused as possible, so I would just merge the changes or merge conflicts from the main branch on the other PR.

andresmgot: gotcha, was unclear if this was just for the test. In that case, the preferred way of fixing the accessibility issue (I just learned that) is to use an ID to match the label (so  you don't have to repeat the label and aria-label). You can see how that is done here with a combination of adding `htmlFor` in the `label` (`InlineField` here) and `inputID` in the select:

https://github.com/grafana/grafana/pull/44277/files#diff-02eddbfa5536d3320579939f0990a9af2835adc752bd5022a1ccce502287a073R27

> I prefer to keep PRs as small and as focused as possible, so I would just merge the changes or merge conflicts from the main branch on the other PR.

yep, totally agree

andresmgot (approved):
:+1:

PR #309, opened 2022-01-27
Add import dashboard e2e scenario

### LINKED ISSUE #241: Add E2E tests for ADX
Use importDashboards flow from main grafana to smoke test ADX to start. May require upgrading some packages to get the latest from @grafana/e2e in there. Some of this work was started and then put on hold, I think it might make sense to just start over with this work. Going to close out some of the old tickets since they're just kind of stale now.

fridgepoet:
Adds a dashboard from a JSON (2 panels: one timeseries and one table view)

<s>Notes for reviewer:
* I don't really understand the expected use of `describe` vs `it` in the @grafana/e2e package. Having a talk with Vicky about that in a few hours...
* It seems like importing dashboard passes *without* adding the datasource first. (For example running this test alone.) Does that sound right?
* There is always a `temporary` test in the cypress UI. I think this might be coming from the @grafana/e2e `after` each test flow?</s>

Closes #241

### DIFF
--- a/cypress/dashboards/example.json
+++ b/cypress/dashboards/example.json
@@ -22,12 +22,12 @@
   "fiscalYearStartMonth": 0,
   "gnetId": null,
   "graphTooltip": 0,
-  "id": 2,
+  "id": null,
   "links": [],
   "liveNow": false,
   "panels": [
     {
-      "datasource": null,
+      "datasource": "e2e-azure-data-explorer-datasource",
       "fieldConfig": {
         "defaults": {
           "color": {
@@ -113,9 +113,9 @@
             }
           },
           "pluginVersion": "3.6.1",
-          "query": "PerfTest\n| where $__timeFilter(_Timestamp_)\n| order by _Timestamp_ asc",
-          "querySource": "visual",
-          "rawMode": false,
+          "query": "PerfTest\n| limit 10\n| order by _Timestamp_ asc",
+          "querySource": "raw",
+          "rawMode": true,
           "refId": "A",
           "resultFormat": "time_series"
         }
@@ -124,7 +124,7 @@
       "type": "timeseries"
     },
     {
-      "datasource": null,
+      "datasource": "e2e-azure-data-explorer-datasource",
       "fieldConfig": {
         "defaults": {
           "color": {
@@ -180,7 +180,7 @@
             }
           },
           "pluginVersion": "3.6.1",
-          "query": "PerfTest\n",
+          "query": "PerfTest\n| limit 10\n| order by _Timestamp_ asc",
           "querySource": "raw",
           "rawMode": true,
           "refId": "A",
@@ -206,5 +206,5 @@
   "timezone": "",
   "title": "Example Dashboard",
   "uid": "EYtcnAx7k",
-  "version": 3
+  "version": 1
 }
--- a/cypress/dashboards/example.json
+++ b/cypress/dashboards/example.json
@@ -1,0 +1,210 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": "-- Grafana --",
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "target": {
+          "limit": 100,
+          "matchAny": false,
+          "tags": [],
+          "type": "dashboard"
+        },
+        "type": "dashboard"
+      }
+    ]
+  },
+  "editable": true,
+  "fiscalYearStartMonth": 0,
+  "gnetId": null,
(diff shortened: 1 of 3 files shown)


cypress/dashboards/example.json
```diff
@@ -0,0 +1,210 @@
...
+    "list": []
+  },
+  "time": {
+    "from": "now-6h",
+    "to": "now"
+  },
+  "timepicker": {},
+  "timezone": "",
+  "title": "Example Dashboard",
+  "uid": "EYtcnAx7k",
```
andresmgot, line 208: you need to remove the uid so there is no possible conflict

cypress/dashboards/example.json
```diff
@@ -0,0 +1,210 @@
...
+  "editable": true,
+  "fiscalYearStartMonth": 0,
+  "gnetId": null,
+  "graphTooltip": 0,
+  "id": 2,
+  "links": [],
+  "liveNow": false,
+  "panels": [
+    {
+      "datasource": null,
```
andresmgot, line 30: here is probably where you need set the datasource info. Something like:

```
"datasource": "e2e-azure-data-explorer-datasource"
```

The problem is that in newer versions, it should be something like:
```
          "datasource": {
            "type": "grafana-azure-monitor-datasource",
            "uid": "Cq2BrVeGz"
          },
```
(and I have no idea how to obtain that `uid`)

But as you discovered, the import does not fail if the datasource doesn't exist. So this test is not really doing anything. I remember that @sarahzinger mentioned that we could extend this test to do something else. You can check with her.

andresmgot:
> * I don't really understand the expected use of `describe` vs `it` in the @grafana/e2e package. Having a talk with Vicky about that in a few hours...

A `describe` includes several tests `it` just one. Not sure what are the steps run after each `describe` (likely some cleanup).

> * It seems like importing dashboard passes _without_ adding the datasource first. (For example running this test alone.) Does that sound right?

Good find, it seems that's how Grafana behaves (and it will select the default data source in that case). We probably need to extend the importDashboard test if we want to be more useful.

> * There is always a `temporary` test in the cypress UI. I think this might be coming from the @grafana/e2e `after` each test flow?

The code at packages/grafana-e2e/src/support/scenario.ts says:

```
      // @todo remove when possible: https://github.com/cypress-io/cypress/issues/2831
      it('temporary', () => {});
```

cypress/dashboards/example.json
fridgepoet: Hmm It sounds like a test adding a dashboard from UI would be more useful for the moment

fridgepoet: Oh also I could also add the datasource at the beginning of this scenario.

cypress/dashboards/example.json
fridgepoet: Interestingly the `Dashboard` type expected by `e2e.flows.importDashboard` requires a uid

cypress/dashboards/example.json
sunker: Yeah since 8.3.0, the datasource name is replaced by a data source ref (with type and uid). See https://github.com/grafana/grafana/pull/33817. I believe this change is forward compatible but not backwards compatible. So if we want to e2e test grafana < 8.3, we should use the data source name as Andrés suggests. Sorry if I'm missing context here, but do you know the name of the data source while importing the dashboard? If so, it should be safe to use it.

sarahzinger: I think if the import doesn't fail if the datasource doesn't exist, maybe that's a bug with importDashboards that we could fix, let me know if you'd like to pair on it!

fridgepoet: Thanks @sarahzinger for pairing with me.

Now that I correctly reference the datasource, it fails as we expect on panel validation if that datasource was not added before.

cypress/integration/smoke.spec.ts
```diff
@@ -19,27 +19,31 @@ type ADXProvision = {
   datasources: ADXConfig[];
 };
 
+function addCommonProvisioningADXDatasource(ADXProvisions: ADXProvision[]) {
+  const datasource = ADXProvisions[0].datasources[1]; // Second entry in our common provisioning file
```
andresmgot, line 23: I'm curious, does the first entry even work? (maybe we can delete it?)

andresmgot (approved):
LGTM! (apart from the issues with the importDashboard flow, but that's nothing to be fixed here)

cypress/integration/smoke.spec.ts
fridgepoet: If we don't use the other entries, I'll delete all of them...

fridgepoet: Anyway the first one doesn't work.

PR #331, opened 2022-03-10
Prepare 3.7.0 stable release

sunker:


andresmgot (approved):
have we received any feedback about the beta?

sunker:
> have we received any feedback about the beta?

Yes ASML has accepted it, so I promised to make stable release.

Issue #224, opened 2021-05-04
Use @grafana/UI CodeEditor


joshhunt:
Once the CodeEditor component in @grafana/UI has kusto language support, we should switch to it :) 

Because we use the version of @grafana/UI from the Grafana instance, we need to keep this in mind for backwards compatibility.

Closes #186 
Closes #139

vickyyyyyyy:
done in https://github.com/grafana/azure-data-explorer-datasource/pull/276

andresmgot:
Had to revert #276 so reopening this. Take into account the issues reported at #284 when retrying this.

Issue #325, opened 2022-02-24
Edit KQL button does not display a panel letting me edit KQL


robeving:
**What happened**:
I have successfully install the plugin and querying my data with the builder interface. I now want to use much more complex KQL queries which require me to use a custom query. I am expecting to click the Edit KQL button and be presented with a box to enter whatever text I like. I press the button and the UI creates a large area under the button but this area does not contain a text box.

**What you expected to happen**:
Text box appears to enter KQL.

**How to reproduce it (as minimally and precisely as possible)**:
Click Edit KQL button

**Screenshots**
![image](https://user-images.githubusercontent.com/38658282/155549927-a5b846d3-15e1-4bb6-a3ff-86b310015935.png)

**Anything else we need to know?**:

**Environment**:

- Grafana version: 8.33
- Plugin version: 3.6.1
- OS Grafana is installed on: Linux
- User OS & Browser: Edge/Chrome
- Others:

sunker:
Thanks for creating this issue @robeving. 

I'm afraid I'm not able to reproduce the problem. Would you be able to check for errors in the javascript console and report back here if any?

andresmgot:
I cannot find a reference but this is a known issue. If the Monaco editor bundled with Grafana is loaded and then you try to use the one bundled by ADX you get the error: `Uncaught TypeError: Cannot read properties of undefined (reading 'kustoDefaults')` and the code editor doesn't load. As a workaround, you can set up ADX as the default data source so no other code editor is loaded before ADX.

This will be fixed once we work on  #288

taleena:
@andresmgot is this now resolved given the linked issues are now closed, and which release is this in?

vickyyyyyyy:
I'm guessing we need a new release to go out with the new editor changes? The last release was in March: https://github.com/grafana/azure-data-explorer-datasource/releases/tag/v3.7.1 and it looks like the new editor (which should resolve this) has been worked on since?

andresmgot:
yes, we need to release a new major version (4.0) so we need to think the best timing for it

taleena:
This is released as of today.

hfleitas:
Happening in Azure Managed Grafana 8.4.5, targets 3.7.1

PR #342, opened 2022-04-01
e2e : Added panel creation scenario

### LINKED ISSUE #313: Add e2e scenario running queries
### Ideas for scenarios
*  In our common provisioned datasource, there seems to be some data present in PerfTest between 2019-08-01 to 2019-10-31
  * In one scenario, could type the KQL editor: 
    ```
    PerfTest
    | where $__timeFilter(_Timestamp_)
    | order by _Timestamp_ asc
    ```
  * In another scenario, could create the same panel as above using the UI.
* What should we assert/validate appears?

### "Worries"
* Using "real" data is brittle/fragile. 
  * That datasource could become unavailable (already a problem for all other tests actually)
  * The data could be changed/erased

yaelleC:
Fixes #313 

Added two e2e scenarios:

1. Types in a KQL query (`PerfTest | where $__timeFilter(_Timestamp_) | order by _Timestamp_ asc`) and runs it
      -> check that panel loads and some data is shown (looking for flag: `Lonely period range deg`)
      

      
2. Uses the builder to run a query returning no results (looking at `events.all` table)
      -> check that the panel display message : `No graphable fields`

### DIFF
--- a/cypress/integration/smoke.spec.ts
+++ b/cypress/integration/smoke.spec.ts
@@ -37,6 +37,18 @@ function addCommonProvisioningADXDatasource(ADXProvisions: ADXProvision[]) {
 }
 
 e2e.scenario({
+  describeName: 'Add ADX datasource',
+  itName: 'fills out datasource connection configuration',
+  scenario: () => {
+    e2e()
+      .readProvisions(['datasources/adx.yaml'])
+      .then((ADXProvisions: ADXProvision[]) => {
+        addCommonProvisioningADXDatasource(ADXProvisions);
+      });
+  },
+});
+
+e2e.scenario({
   describeName: 'Import dashboard',
   itName: 'fills out datasource connection configuration, imports JSON dashboard, adds panel and run query',
   scenario: () => {
--- a/cypress/integration/smoke.spec.ts
+++ b/cypress/integration/smoke.spec.ts
@@ -75,12 +75,49 @@ e2e.scenario({
             e2eSelectors.queryEditor.database.input().click({ force: true });
             cy.contains('PerfTest').click();
             e2eSelectors.queryEditor.codeEditorLegacy
-              .container()
-              .click({ force: true })
-              .type(`{selectall} {backspace} PerfTest | where $__timeFilter(_Timestamp_) | order by _Timestamp_ asc`);
+              .textarea()
+              .should('be.visible')
+              .clear()
+              .click()
+              .type(`PerfTest | where $__timeFilter(_Timestamp_) | order by _Timestamp_ asc`);
             e2eSelectors.queryEditor.runQuery.button().click({ force: true });
             cy.get('.panel-loading');
             cy.get('.panel-loading', { timeout: 10000 }).should('not.exist');
+          },
+        });
+      });
+  },
+});
+
+e2e.scenario({
+  describeName: 'Creates Panel run query via builder',
+  itName: 'fills out datasource connection configuration, adds panel and runs query via builder',
+  scenario: () => {
+    e2e()
+      .readProvisions(['datasources/adx.yaml'])
+      .then((ADXProvisions: ADXProvision[]) => {
+        addCommonProvisioningADXDatasource(ADXProvisions);
+
+        e2e.flows.addDashboard({
+          timeRange: {
+            from: '2022-01-05 19:00:00',
+            to: '2022-01-10 19:00:00',
+          },
+          variables: [],
+        });
+
+        e2e.flows.addPanel({
+          matchScreenshot: false,
+          visitDashboardAtStart: false,
+          queriesForm: () => {
+            e2eSelectors.queryEditor.database.input().click({ force: true });
+            cy.contains('PerfTest').click();
+
+            e2eSelectors.queryEditor.tableFrom.input().click({ force: true });
+            cy.contains('events.all').click();
+
+            e2eSelectors.queryEditor.runQuery.button().click({ force: true });
+            cy.contains('No graphable fields').should('exist');
           },
         });
       });
--- a/cypress/integration/smoke.spec.ts
+++ b/cypress/integration/smoke.spec.ts
@@ -71,18 +71,17 @@ e2e.scenario({
           matchScreenshot: false,
           visitDashboardAtStart: false,
           queriesForm: () => {
-            e2eSelectors.queryEditor.editKQL.button().click({ force: true });
             e2eSelectors.queryEditor.database.input().click({ force: true });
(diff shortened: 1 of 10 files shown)


cypress/integration/smoke.spec.ts
```diff
@@ -50,14 +50,23 @@ e2e.scenario({
...
       .then((ADXProvisions: ADXProvision[]) => {
         addCommonProvisioningADXDatasource(ADXProvisions);
 
         e2e.flows.importDashboard(TEST_DASHBOARD);
+
+        e2e.flows.addPanel({
+          matchScreenshot: false,
+          visitDashboardAtStart: false,
+          queriesForm: () => {
+            e2eSelectors.queryEditor.editKQL.button().click({ timeout: 5000 });
```
yaelleC, line 66: Somehow I cannot access the monaco editor in e2e tests - getting:

Screenshot 2022-04-04 at 12 16 16

andresmgot: you hit https://github.com/grafana/azure-data-explorer-datasource/issues/325 :) be sure you don't load any monaco editor prior to this

yaelleC: Thanks! Not sure how I can enforce that though, I suppose the other monaco editor is the one used in dashboard upload json, I will try to split the tests.

Otherwise, I am currently working on a workaround that uses the builder instead of the KQL editor...

andresmgot: > I suppose the other monaco editor is the one used in dashboard upload json, I will try to split the tests.

I remember that situation triggering the bug so yes. If it's not possible to avoid that, I would disable the JSON upload test while we fix #224, it's less valuable than the one using the editor.

yaelleC: I just moved it to a separate test and that works fine, thanks!

cypress/integration/smoke.spec.ts
```diff
@@ -34,30 +35,54 @@ function addCommonProvisioningADXDatasource(ADXProvisions: ADXProvision[]) {
...
 e2e.scenario({
   describeName: 'Add ADX datasource',
-  itName: 'fills out datasource connection configuration',
+  itName: 'fills out datasource connection configuration and imports JSON dashboard',
   scenario: () => {
     e2e()
       .readProvisions(['datasources/adx.yaml'])
       .then((ADXProvisions: ADXProvision[]) => {
         addCommonProvisioningADXDatasource(ADXProvisions);
+        e2e.flows.importDashboard(TEST_DASHBOARD);
```
yaelleC, line 48: moved here to avoid having 2 monaco editor open in same test (bug #325)

cypress/integration/smoke.spec.ts
```diff
@@ -34,30 +35,54 @@ function addCommonProvisioningADXDatasource(ADXProvisions: ADXProvision[]) {
...
       .readProvisions(['datasources/adx.yaml'])
       .then((ADXProvisions: ADXProvision[]) => {
         addCommonProvisioningADXDatasource(ADXProvisions);
 
-        e2e.flows.importDashboard(TEST_DASHBOARD);
+        e2e.flows.addDashboard({
+          timeRange: {
+            from: '2019-10-05 19:00:00',
+            to: '2019-10-10 19:00:00',
+          },
```
yaelleC, line 66: attow there is data for this time period in `PerfTest`

cypress/integration/smoke.spec.ts
```diff
@@ -34,30 +35,91 @@ function addCommonProvisioningADXDatasource(ADXProvisions: ADXProvision[]) {
...
+              .should('be.visible')
+              .clear()
+              .click()
+              .type(`PerfTest | where $__timeFilter(_Timestamp_) | order by _Timestamp_ asc`);
+            e2eSelectors.queryEditor.runQuery.button().click({ force: true });
+            cy.get('.panel-loading');
+            cy.get('.panel-loading', { timeout: 10000 }).should('not.exist');
+          },
+        });
+      });
```
andresmgot, line 88: Tried this and this scenario was the one failing for me. In particular `clear` was not clearing the textarea (probably something weird in our component). Also, `.panel-loading' was not being rendered either. I modified those two and this seems to work for me:

```suggestion
        e2e.flows.addPanel({
          matchScreenshot: false,
          visitDashboardAtStart: false,
          queriesForm: () => {
            e2eSelectors.queryEditor.database.input().click({ force: true });
            cy.contains('PerfTest').click();
            e2eSelectors.queryEditor.editKQL.button().click({ force: true });
            e2eSelectors.queryEditor.codeEditorLegacy
              .textarea()
              .should('be.visible')
              .click({ force: true })
              .type('{del}'.repeat(100))
              .type(`PerfTest | where $__timeFilter(_Timestamp_) | order by _Timestamp_ asc`);
            e2eSelectors.queryEditor.runQuery.button().click({ force: true });
            cy.contains('Lonely period range deg');
          },
        });
      });
```

hope that helps!

yaelleC:
The `RawQueryEditor.test.tsx` and `RawQueryEditor.tsx` changes are not needed anymore (using `data-test-id` from `selectors`) but I feel like they might become useful in the future (when adding e2e tests for the new editor) so decided to leave them there.
Happy to remove them to simplify the PR if need be.

andresmgot (approved):
LGTM!

src/components/VisualQueryEditor.tsx
```diff
@@ -213,16 +214,23 @@ export const VisualQueryEditor: React.FC<Props> = (props) => {
...
+      templateVariableOptions={props.templateVariableOptions}
+      label="From"
+      value={table}
+      fields={tables}
+      onChange={onChangeTable}
+      aria-label={selectors.components.queryEditor.tableFrom.input}
+    >
+      {kustoTableProps.children}
+    </KustoPropertyEditorSection>
+  );
```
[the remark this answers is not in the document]
yaelleC: I did not want to have to add the `aria-label` everywhere 😉 
But I could settle for a const and repeat the `aria-label` twice if that feels less weird?
```
  const KustoTable = 
    <KustoPropertyEditorSection
      templateVariableOptions={props.templateVariableOptions}
      label="From"
      value={table}
      fields={tables}
      onChange={onChangeTable}
      aria-label={selectors.components.queryEditor.tableFrom.input}
   />
```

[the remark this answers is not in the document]
andresmgot: my brain is already a bit off today so I am not sure what you mean by twice? but in any case, it's not better how it's at the moment so feel free to merge it as you have it.

PR #345, opened 2022-04-01
Fix CSS import

### LINKED ISSUE #343: Unable to run tests on files importing css from 'emotion'
**What happened**:

Received an error:

```
"emotion" has been renamed to "@emotion/css", please import it from "@emotion/css" instead
```

So we need to update those import paths (and check if this breaks compatibility with Grafana 7.5).

**Screenshots**

**Environment**:

- Grafana version:
- Plugin version:
- OS Grafana is installed on:
- User OS & Browser:
- Others:

andresmgot:
Fixes #343

Fixes import path for the `css` package. Note that this breaks compatibility with Grafana 7.5 since it uses an old version of the package. That's why I have updated the changelog for a future major version release.

@taleena can you confirm if this is okay?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## [4.0.0]
+
+_Not released yet_
+
+- Breaking Change: Azure Data Explorer plugin now requires Grafana 8.0+ to run.
+
 ## [3.7.1]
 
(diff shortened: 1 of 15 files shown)


yaelleC (approved):

taleena:
Confirmed in person, mentioning here just in case.

PR #346, opened 2022-04-01
Remove dead code that references to Angular

### LINKED ISSUE #288: Remove angular

andresmgot:
It turns out that the only way where we were using angular was in the annotation query editor but that editor was removed in #265 so the code was unreachable.

I checked the editor and it's the regular query editor (and I was able to run some example):

Screenshot from 2022-04-01 17-00-35

Can you suggest a working annotation query here? (so I can see that it really works).

Closes #288


yaelleC (approved):
Tested locally with the following and works like a charm 👍 
Both with builder and with KQL editor

```
PerfTest
| where $__timeFilter(_Timestamp_)
| project _Timestamp_, _val1_
| order by _Timestamp_ asc
```

Screenshot 2022-04-04 at 10 06 21

sunker (approved):
Annotations seems to work after this change. You can test it yourself with the attached dashbaord.
[ADX Annotations-1649062123412.zip](https://github.com/grafana/azure-data-explorer-datasource/files/8408191/ADX.Annotations-1649062123412.zip)

I believe we should be able to remove the config_editor.ts too.

andresmgot:
thanks for confirming!

PR #347, opened 2022-04-04
Replace deprecated stylesFactory with useStyles2

andresmgot:
- Replace deprecated stylesFactory with useStyles2
- Migrate src/components/RawQueryEditor.tsx to a function component (so it can use useStyles2)
- Migrate src/editor/components/filter/QueryEditorFieldAndOperator.tsx to a function component (so it can use useStyles2)
- Migrate src/editor/components/operators/QueryEditorOperator.tsx to a function component (so it can use useStyles2)

There is still a file using stylesFactory: src/monaco/KustoMonacoEditor.tsx but that file should be removed after #224

No functional changes here.


andresmgot:
okay, need to check the CI issues first

andresmgot:
Ready to be reviewed ^

sunker (approved):
LGTM

Issue #284, opened 2021-11-23
KQL Editor broken 


pahansen:
**What happened**:
Since today we are not able to use the KQL query editor anymore. It is only possible to query data when using the "Builder". Changes can not be saved in KQL editor, tables are not detected, "Run Query" button does not trigger searches and intellisense is not working anymore. 

Additionally, the "Builder" also breaks when going back and forth between Builder and KQL Editor.

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:
Use raw Query in KQL editor and try to edit query.

**Screenshots**
![image](https://user-images.githubusercontent.com/39408669/143039751-6e93689a-c4d9-482b-8e7d-4b5a8b0bee33.png)
![image](https://user-images.githubusercontent.com/39408669/143039853-e1642e59-8996-4293-8017-2a937bca18e9.png)
![image](https://user-images.githubusercontent.com/39408669/143040172-3a7e6db7-1b0d-43a5-a31d-eabf33e162d9.png)

**Anything else we need to know?**:

**Environment**:
- Grafana version: 8.0.6
- Plugin version: 3.6.0
- OS Grafana is installed on: Docker (locally)
- User OS & Browser: Chrome
- Others:

FVE-DUCO:
I have the same!
Environment:

    Grafana version: 8.2.5 
    Plugin version: 3.6.0
    OS Grafana is installed on Linux
    User OS & Browser: Windows, Firefox
    Others:

andresmgot:
Hi,

Thanks for reporting the issue. This is related to https://github.com/grafana/azure-data-explorer-datasource/pull/276 that was just released in 3.6.0. I am going to take a look to the issue but in the meantime, you can downgrade to the previous version (3.5.1). Sorrry for the inconveniences!

andresmgot:
@FVE-DUCO @pahansen released 3.6.1 with the same editor than before. That should solve your issues.

cdeil:
We also ran into this issue today. Not sure if it helps, but FWIW this is the query error we got:
```
Failed to execute query: Azure HTTP "520 InternalServiceError": "name\r\nParameter name: Argument 'name' is empty: at .ctor in C:\\source\\Src\\Engine\\DataNode\\Common\\Entities\\DatabaseIdentifier.cs: line 87
```

shacharz:
I ran into the same problem, with plugin version 3.6.1.

```
Grafana version: 8.2.5
[...]

PR #348, opened 2022-04-05
Add @grafana/ui CodeEditor behind a feature flag

andresmgot:
Starting point for #224 

This PR adds the new code editor behind a feature flag. Right now the editor has no additional functionality (apart from syntax highlighting) but those can be added incrementally without affecting possible releases.

I have also added some basic tests and mocks.

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -6,7 +6,7 @@ import { QueryEditor } from './QueryEditor';
 import { AdxDataSourceOptions, AdxDataSourceSecureOptions, KustoQuery } from './types';
 
 export const plugin = new DataSourcePlugin<AdxDataSource, KustoQuery, AdxDataSourceOptions, AdxDataSourceSecureOptions>(
-  AdxDataSource as any
+  AdxDataSource
 )
   .setConfigEditor(ConfigEditor)
   .setQueryEditor(QueryEditor);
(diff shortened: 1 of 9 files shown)


src/module.ts
```diff
@@ -1,14 +1,12 @@
...
-import ConfigEditor from 'components/ConfigEditor';
 
-export const plugin = new DataSourcePlugin<
-  DataSourceApi<KustoQuery, AdxDataSourceOptions>,
-  KustoQuery,
-  AdxDataSourceOptions,
-  AdxDataSourceSecureOptions
->(AdxDataSource as any)
+export const plugin = new DataSourcePlugin<AdxDataSource, KustoQuery, AdxDataSourceOptions, AdxDataSourceSecureOptions>(
+  AdxDataSource as any
```
sunker, line 9: Remove `as any`?

sunker (approved):
Nice!

src/module.ts
sunker, line 9: Remove `as any`?

sunker (approved):
Nice! Do we have an idea of what kind of issues we'll be running into when using the CodeEditor? Were these issues closed when we reverted back to the legacy editor a while back?

yaelleC (approved):
LGTM
That new editor would make my e2e tests a lot easier to set up 😉

src/module.ts
andresmgot: missed this. Thanks!

andresmgot:
> Nice! Do we have an idea of what kind of issues we'll be running into when using the CodeEditor? Were these issues closed when we reverted back to the legacy editor a while back?

Yes, explained here: https://github.com/grafana/azure-data-explorer-datasource/issues/284. Once I investigate that more I will sanitize the feedback in bullet points in https://github.com/grafana/azure-data-explorer-datasource/issues/224

PR #350, opened 2022-04-06
Add macro suggestions

andresmgot:
Another incremental change towards #224

In this PR, I am adding the data source macros as suggestions. The legacy implementation was smart enough to suggest different macros depending on the line content so this PR does the same. It's not possible to use `getSuggestions` from the `CodeEditor` since that return "dumb" suggestions (always the same ones). Because of that, we need to access the underlying Monaco editor in the `onEditorDidMount` function.

The legacy code for this can be found here: https://github.com/grafana/azure-data-explorer-datasource/blob/main/src/monaco/kusto_code_editor.ts#L167

And the tests are the same than for the legacy editor, which can be found here: https://github.com/grafana/azure-data-explorer-datasource/blob/main/src/monaco/kusto_code_editor.test.ts#L7

Peek 2022-04-06 17-59

### DIFF
--- a/src/components/RawQueryEditor.tsx
+++ b/src/components/RawQueryEditor.tsx
@@ -59,16 +59,16 @@ export const RawQueryEditor: React.FC<RawQueryEditorProps> = (props) => {
 
   const styles = useStyles2(getStyles);
 
-  if (!schema) {
-    return null;
-  }
-
   const handleEditorMount = useCallback((editor: MonacoEditor, monaco: Monaco) => {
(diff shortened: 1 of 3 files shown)


yaelleC (approved):
lgtm, 

I was wondering: do we have any throttling in place for autocomplete / suggestions?

andresmgot:
> I was wondering: do we have any throttling in place for autocomplete / suggestions?

I am not sure what you mean?

andresmgot:
merging this to send the next PR, will address future comments if needed

PR #352, opened 2022-04-07
Give signature suggestions

andresmgot:
The signature suggestions are given when some function parameters need to be specified (it's only triggered when opening parenthesis). In this case, to maintain previous behavior, we are giving some information when the user tries to use the `$__timeFilter` macro:

Screenshot from 2022-04-07 17-50-54

### DIFF
--- a/src/components/Suggestions.ts
+++ b/src/components/Suggestions.ts
@@ -4,18 +4,18 @@ import { includes } from 'lodash';
 const defaultTimeField = 'Timestamp';
 
 export function getSuggestions(model: monacoTypes.editor.ITextModel, position: monacoTypes.Position) {
-  const word = model.getWordUntilPosition(position);
+  const wordPosition = model.getWordUntilPosition(position);
   const prevChar = model.getValueInRange({
     startLineNumber: position.lineNumber,
     endLineNumber: position.lineNumber,
-    startColumn: word.startColumn - 1,
-    endColumn: word.startColumn,
+    startColumn: wordPosition.startColumn - 1,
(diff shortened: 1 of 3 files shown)


src/components/Suggestions.ts
```diff
@@ -105,3 +105,41 @@ export function getSuggestions(model: monacoTypes.editor.ITextModel, position: m
     suggestions: [],
   };
 }
+
+export function getSignatureHelp(
+  model: monacoTypes.editor.ITextModel,
+  position: monacoTypes.Position
+): monacoTypes.languages.ProviderResult<monacoTypes.languages.SignatureHelpResult> {
+  const word = model.getWordUntilPosition(position.delta(0, -1));
```
sarahzinger, line 113: hi! mind talking through what this is doing I'm not sure I follow why we need word or what word this represents and how this is different from textUntilPosition? Couldn't we just pass 0 for the startColumn or am I misunderstanding?

sarahzinger (approved):
neat! I had one question for my own clarification, but I think it should be fair to ship :) maybe we should add it to changelog.

src/components/Suggestions.ts
andresmgot: true,  `word` may be misleading, this is actually the `wordPosition`. I just noticed that a word position actually has the property `word` (which is the word text) so I can highly simplify this :)

andresmgot:
> maybe we should add it to changelog.

In theory, there should be "no changes" from the user perspective. I am just migrating functionality from the legacy editor to the new one (without adding anything new so far).

PR #355, opened 2022-04-13
Load database schema

andresmgot:
Cont #224 

Load the database schema so we get more suggestions like table names and columns:

Screenshot from 2022-04-13 16-59-46

Next step would be to add macros and templates variables as done for [Azure Monitor Logs](https://github.com/grafana/grafana/pull/47689).

Unfortunately, this will only work with the latest version of Grafana (>=8.5.0). This is because only that version includes the latest version of `@kusto/monaco-kusto`, which is needed to successfully load the schema. Once the editor is ready and we remove the feature flag, we can substitute it for a version check and only load the new editor if the Grafana version is bigger than 8.5.0 (:unamused:)

Note: It's not possible to test this with unit tests (it's not possible to load languages with jest). My plan is to add a check in the e2e test to verify this but will do that in a different PR.

### DIFF
--- a/src/components/RawQueryEditor.tsx
+++ b/src/components/RawQueryEditor.tsx
@@ -4,11 +4,11 @@ import { config } from '@grafana/runtime';
 import { CodeEditor, Icon, Monaco, MonacoEditor, useStyles2 } from '@grafana/ui';
 import { QueryEditorResultFormat, selectResultFormat } from 'components/QueryEditorResultFormat';
 import { AdxDataSource } from 'datasource';
+import { KustoMonacoEditor } from 'monaco/KustoMonacoEditor';
 import React, { useCallback, useState } from 'react';
 import { selectors } from 'test/selectors';
 import { AdxDataSourceOptions, AdxSchema, KustoQuery } from 'types';
 
-import { KustoMonacoEditor } from '../monaco/KustoMonacoEditor';
 import { getSignatureHelp, getSuggestions } from './Suggestions';
 
 type Props = QueryEditorProps<AdxDataSource, KustoQuery, AdxDataSourceOptions>;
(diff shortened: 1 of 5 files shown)


src/components/RawQueryEditor.tsx
```diff
@@ -60,7 +60,7 @@ export const RawQueryEditor: React.FC<RawQueryEditorProps> = (props) => {
 
   const styles = useStyles2(getStyles);
 
-  const handleEditorMount = useCallback((editor: MonacoEditor, monaco: Monaco) => {
+  const handleEditorMount = useCallback(async (editor: MonacoEditor, monaco: Monaco) => {
```
sarahzinger, line 63: I don't think this `async` does anything if you don't have `await`

src/types.ts
```diff
@@ -125,6 +126,8 @@ export interface AdxTableSchema {
 export interface AdxColumnSchema {
   Name: string;
   CslType: string;
+  Type?: string;
```
sarahzinger, line 129: can you talk about these changes a little more? I'm not sure I see where they are being used in this pr

sarahzinger (approved):
looks good, I agree an e2e test would be best in this case if possible, since any unit tests here will be too mocky to be useful.

src/components/RawQueryEditor.tsx
```diff
@@ -69,6 +69,17 @@ export const RawQueryEditor: React.FC<RawQueryEditorProps> = (props) => {
       signatureHelpTriggerCharacters: ['(', ')'],
       provideSignatureHelp: getSignatureHelp,
     });
+    monaco.languages['kusto']
```
kevinwcyu, line 72: This should be returned if the caller of `handleEditorMount` is expecting a promise.

kevinwcyu (approved):

src/components/RawQueryEditor.tsx
andresmgot: right, I forgot to remove this.

src/components/RawQueryEditor.tsx
andresmgot: not really needed to return a promise.

src/types.ts
andresmgot: not needed (yet), I forgot to remove this in this PR.

Issue #356, opened 2022-04-13
Generalize generation of scopes for all clouds


kostrse:
Currently, US Government cloud uses cluster URL as a scope and other clouds have predefined scope strings.

This can be generalized for all clouds:
1. All clouds may have the same logic based on the subdomain of cluster URL.
2. Proper top level domains for a given cloud (`*.windows.net`, `*.chinacloudapi.cn`) should be enforced. Requesting arbitrary scope based on free input of cluster URL should be discourage.

fridgepoet:
> Requesting arbitrary scope based on free input of cluster URL should be discourage.

So were you envisioning some kind of verification on `clusterUrl`? Or something where we wanted to be sure that the `clusterUrl` passed in would never be a free input? Or did you have something else in mind?

fridgepoet:
What would you envision for the errors returned by [normalizeAzureCloud](https://github.com/grafana/azure-data-explorer-datasource/pull/358/files#diff-4945be9fedff9efd9dc862ba5a30cfb155c5fc2b109030d838d05415c6c30c5fR16) and [resolveAuthorityForCloud](https://github.com/grafana/azure-data-explorer-datasource/pull/358/files#diff-27b85ab823a85d1aa0ba5d42b38e77909835a2e2f6e3ce9a6eaccc482abccc7fR60)?

sunker:
Sounds good, but the tricky part here is testing. It took us months to get [this](https://github.com/grafana/azure-data-explorer-datasource/pull/258) tested in Mooncake cloud and in the USGov cloud.

aangelisc:
@kostrse is this still an issue? Since we've added endpoint enforcement

PR #358, opened 2022-04-13
Migrate client secret auth to Grafana Azure SDK

kostrse:
Introduced dependency on the external module [github.com/grafana/grafana-azure-sdk-go](https://github.com/grafana/grafana-azure-sdk-go).

App Registration (client secret) authentication was migrated to the library. OBO still implemented in ADX because it's not supported by the library.


kostrse:
cc @sunker @andresmgot

pkg/azuredx/azureauth/scopes.go
```diff
@@ -0,0 +1,35 @@
...
+
+var (
+	azureDataExplorerScopes = map[string][]string{
+		azsettings.AzurePublic:       {"https://kusto.kusto.windows.net/.default"},
+		azsettings.AzureChina:        {"https://kusto.kusto.chinacloudapi.cn/.default"},
+		azsettings.AzureUSGovernment: {"{clusterUrl}/.default"},
+	}
+)
+
+func getAzureScopes(credentials *azcredentials.AzureClientSecretCredentials, clusterUrl string) ([]string, error) {
```
fridgepoet, line 19: In this function, could you have addressed TODO #356 by checking clusterUrl for special case top domains for `AzurePublic` and `AzureChina` and if neither of those, then simply applying the generalized substitution 
`[]string{fmt.Sprintf("%s/.default", clusterUrl)}`?

Or does that not protect against the free input of cluster URL as you mentioned?

pkg/azuredx/azureauth/auth.go
```diff
@@ -53,55 +35,74 @@ func init() {
...
-	HTTPDo func(req *http.Request) (*http.Response, error)
-	// ServicePrincipalToken is the azidentity.ClientSecretCredential GetToken method.
-	ServicePrincipalToken func(ctx context.Context, opts azcore.TokenRequestOptions) (*azcore.AccessToken, error)
-	tokenCache            *cache
-	scope                 string
+	HTTPDo        func(req *http.Request) (*http.Response, error)
+	authority     azidentity.AuthorityHost
+	tokenProvider aztokenprovider.AzureTokenProvider
+	tokenCache    *cache
+	scopes        []string
```
fridgepoet, line 50: How come scopes is an array now? `getAzureScopes` doesn't return more than one scope.

(I see you had mentioned TODO #356 there, but didn't understand from the description of that ticket that more than one scope would be returned either.)

fridgepoet:
(Asked a couple of questions about scope since I saw #356. Sorry if you might have discussed these with someone else on the team or if these are obvious for you.)

pkg/azuredx/azureauth/scopes.go
kostrse: This needs additional testing, I just don't want to address it in this PR to not cause unwanted regressions on top of the significant changes I made. I preserved the original logic and created task to address it separately.

pkg/azuredx/azureauth/auth.go
kostrse: While in this particular case ADX endpoint needs only a single scope (`.default` - is the default, catch-all scope), generally it's not always the case. An endpoint in theory _may_ require more than one scope at the same time, and access token can be issued for more than one scope.

Signature of the `GetAccessToken` func in `AzureTokenProvider` ([here](https://github.com/grafana/grafana-azure-sdk-go/blob/e079ca81c3fb6b309b37843345cf97f30aad4609/aztokenprovider/token_provider.go#L20)):

```go
type AzureTokenProvider interface {
	GetAccessToken(ctx context.Context, scopes []string) (string, error)
}
```

So I made `getAzureScopes` to return `[]string` array more for convention, as a type for representation of scopes, proper abstraction, rather than because it's technically required. The fact that `getAzureScopes` returns only one scope is implementation detail.

pkg/azuredx/azureauth/scopes.go
fridgepoet: Got it.

>I preserved the original logic

It looks like the original `getCloudScope` did not return [this error case for unknown Azure cloud on line 24](https://github.com/grafana/azure-data-explorer-datasource/pull/358/files#diff-98f65fb6bc9c4660e45ec0766e5bf93e54d5cd21eb6443db20f18105a5c5c173R24) unless I'm missing something.

pkg/azuredx/azureauth/scopes.go
```diff
@@ -0,0 +1,35 @@
...
+	// Get scopes for the given cloud
+	if scopesTmpl, ok := azureDataExplorerScopes[azureCloud]; !ok {
+		err := fmt.Errorf("the Azure cloud '%s' not supported by Azure Data Explorer datasource", azureCloud)
+		return nil, err
+	} else {
+		// TODO: #356 Generalize generation of scopes for all clouds
+		clusterUrl = strings.TrimSuffix(clusterUrl, "/")
+		scopes := make([]string, len(scopesTmpl))
+		for i, scopeTmpl := range scopesTmpl {
+			scopes[i] = strings.Replace(scopeTmpl, "{clusterUrl}", clusterUrl, 1)
```
fridgepoet, line 31: > While in this particular case ADX endpoint needs only a single scope (`.default` - is the default, catch-all scope), generally it's not always the case. An endpoint in theory _may_ require more than one scope at the same time, and access token can be issued for more than one scope.
> 
> Signature of the `GetAccessToken` func in `AzureTokenProvider` ([here](https://github.com/grafana/grafana-azure-sdk-go/blob/e079ca81c3fb6b309b37843345cf97f30aad4609/aztokenprovider/token_provider.go#L20)):
> 
> ```go
> type AzureTokenProvider interface {
> 	GetAccessToken(ctx context.Context, scopes []string) (string, error)
> }
> ```
> 
> So I made `getAzureScopes` to return `[]string` array more for convention, as a type for representation of scopes, proper abstraction, rather than because it's technically required. The fact that `getAzureScopes` returns only one scope is implementation detail.



If ADX only ever will return one scope, just for this ADX implementation, what do you think if we returned only `string` and reduced the complexity from the loop here?
(or return the one string wrapped in `[]string{}`)

something along the lines of:

```
var (
	azureDataExplorerScopes = map[string]string{
		azsettings.AzurePublic:       "https://kusto.kusto.windows.net/.default",
		azsettings.AzureChina:        "https://kusto.kusto.chinacloudapi.cn/.default",
		azsettings.AzureUSGovernment: "{clusterUrl}/.default",
	}
)

func getAzureScopes(credentials *azcredentials.AzureClientSecretCredentials, clusterUrl string) ([]string, error) {
	azureCloud := credentials.AzureCloud

	// Get scopes for the given cloud
	scopeTmpl, ok := azureDataExplorerScopes[azureCloud]
	if !ok {
		return nil, fmt.Errorf("the Azure cloud '%s' not supported by Azure Data Explorer datasource", azureCloud)
	}

	return []string{strings.Replace(scopeTmpl, "{clusterUrl}", strings.TrimSuffix(clusterUrl, "/"), 1)}, nil
}

```

kostrse: Yes, makes sense to just wrap the string. I will update the code.

kostrse: Fixed

pkg/azuredx/azureauth/scopes.go
kostrse: I meant I preserved the scope values as they were before: Public/China - predefined string, US Gov - cluster URL.
I added error because falling back to public cloud value doesn't make sense, it won't work, and inconsistent with other parts which have logic based on selected cloud ([here](https://github.com/grafana/grafana-azure-sdk-go/blob/e079ca81c3fb6b309b37843345cf97f30aad4609/aztokenprovider/token_provider.go#L124) and  [here](https://github.com/kostrse/azure-data-explorer-datasource/blob/2076be1224c340329c2a3a3fa1cdc70b6d78fd0d/pkg/azuredx/azureauth/auth.go#L214)).

pkg/azuredx/azureauth/scopes_test.go
```diff
@@ -0,0 +1,55 @@
...
+	for _, tt := range tests {
+		t.Run(tt.description, func(t *testing.T) {
+			scopes, err := getAzureScopes(tt.credentials, tt.clusterUrl)
+			require.NoError(t, err)
+
+			assert.Len(t, scopes, 1)
+			assert.Equal(t, tt.expectedScope, scopes[0])
+		})
+	}
+}
```
fridgepoet, line 55:
```suggestion
}

func TestGetAzureScopes_returns_error_for_unknown_cloud(t *testing.T) {
	_, err := getAzureScopes(&azcredentials.AzureClientSecretCredentials{AzureCloud: "unknown"}, "https://abc.northeurope.kusto.windows.net")

	assert.Error(t, err)
}

```

pkg/azuredx/azureauth/auth.go
```diff
@@ -53,55 +35,74 @@ func init() {
...
 }
 
-func NewServiceCredentials(settings *models.DatasourceSettings, client *http.Client) (*ServiceCredentials, error) {
-	options := &azidentity.ClientSecretCredentialOptions{
-		AuthorityHost: authorityBaseURL(settings.AzureCloud),
+func NewServiceCredentials(settings *models.DatasourceSettings, azureSettings *azsettings.AzureSettings,
+	client *http.Client) (ServiceCredentials, error) {
+	azureCloud, err := normalizeAzureCloud(settings.AzureCloud)
+	if err != nil {
+		return nil, fmt.Errorf("invalid Azure credentials: %s", err)
```
fridgepoet, line 57:
```suggestion
		return nil, fmt.Errorf("invalid Azure credentials: %w", err)
```

pkg/azuredx/azureauth/auth.go
```diff
@@ -53,55 +35,74 @@ func init() {
...
+	}
+
+	authority, err := resolveAuthorityForCloud(azureCloud)
+	if err != nil {
+		return nil, fmt.Errorf("invalid Azure credentials: %s", err)
```
fridgepoet, line 62:
```suggestion
		return nil, fmt.Errorf("invalid Azure credentials: %w", err)
```

pkg/azuredx/azureauth/scopes_test.go
fridgepoet, line 55:
```suggestion
}

func TestGetAzureScopes_returns_error_for_unknown_cloud(t *testing.T) {
	_, err := getAzureScopes(&azcredentials.AzureClientSecretCredentials{AzureCloud: "unknown"}, "https://abc.northeurope.kusto.windows.net")
	assert.Error(t, err)
}

```

pkg/azuredx/azureauth/auth_test.go
```diff
@@ -84,9 +85,13 @@ func TestQueryDataAuthorization(t *testing.T) {
...
 
 		// setup & test
-		c := &ServiceCredentials{HTTPDo: g.HTTPDoMock}
+		authority, err := resolveAuthorityForCloud(azsettings.AzurePublic)
+		require.NoError(t, err)
+
+		c := &ServiceCredentialsImpl{HTTPDo: g.HTTPDoMock}
 		c.TenantID = "0AF0528A-F473-4E0C-891F-3FF8ACDC4E5F"
 		c.OnBehalfOf = true
+		c.authority = authority
```
fridgepoet, line 94:
```suggestion
		c := &ServiceCredentialsImpl{HTTPDo: g.HTTPDoMock}
		c.TenantID = "0AF0528A-F473-4E0C-891F-3FF8ACDC4E5F"
		c.OnBehalfOf = true
		c.authority = azidentity.AzurePublicCloud
```

sunker:
Nice - this is much cleaner than before. Tested with app registration and OBO and it's working as expected. I'll leave the approval of this PR for @fridgepoet.

pkg/azuredx/azureauth/auth_test.go
fridgepoet: Right here I suggest "hardcoding" the `authority` you want in the the setup of your test rather than using a function to return it.

pkg/azuredx/azureauth/scopes_test.go
fridgepoet: Just suggest adding this untested case.

sunker (changes requested):
Actually I'm getting `invalid Azure credentials: the cloud '' not supported`. Seems like ADX was not strict about setting `azureCloud` in the json data, so empty string needs to be treated as public cloud. Can you do that in auth.go?

pkg/azuredx/azureauth/auth.go
kostrse: Thanks, changed

pkg/azuredx/azureauth/scopes_test.go
kostrse: Added

pkg/azuredx/azureauth/auth.go
kostrse: Fixed

kostrse:
> Actually I'm getting `invalid Azure credentials: the cloud '' not supported`. Seems like ADX was not strict about setting `azureCloud` in the json data, so empty string needs to be treated as public cloud. Can you do that in auth.go?

Thanks for pointing at this. Fixed. Unset cloud defaults to public now, any other unknown values fail.

sunker (approved):
Thanks for applying changes! Tested and it's working as expected. 

@fridgepoet looks like all of your comments have been addressed. Is it okay to merge?

fridgepoet (approved):
Yes!

PR #363, opened 2022-04-20
Define macros and variables as functions, not suggestions

andresmgot:
Cont: #224

Before this change, macros and template variables were not interpreted so they were marked as syntax errors and suggested as row text:

Screenshot from 2022-04-20 18-19-07

After this change, those are recognized as functions:

Screenshot from 2022-04-20 18-18-36

### Caveats
I was not able to find fixes for these:

- The only way I found to add template variables is to define them as functions (I tried to do define them as global parameters, but that completely broke autocompletion, with no error reported). This causes that the suggestion is shown with parenthesis. It's fine because when clicking in the suggestion it doesn't add the parenthesis but it can be mildly confusing:
Peek 2022-04-20 18-27

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -126,7 +126,7 @@ export interface AdxTableSchema {
 export interface AdxColumnSchema {
   Name: string;
   CslType: string;
-  Type: string;
+  Type?: string;
   CslDefaultValue?: string;
 }
 
--- a/src/types.ts
+++ b/src/types.ts
@@ -126,6 +126,8 @@ export interface AdxTableSchema {
 export interface AdxColumnSchema {
   Name: string;
   CslType: string;
+  Type: string;
+  CslDefaultValue?: string;
 }
 
 export interface AdxFunctionSchema {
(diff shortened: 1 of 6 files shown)


sunker (commented):
Not saying it's pretty, but can't you try to register two time filter macros? One accepting args and another one that doesn't?

andresmgot:
> Not saying it's pretty, but can't you try to register two time filter macros? One accepting args and another one that doesn't?

I tried that but the schema loading crashes in that case (it is considered a conflict in the argument definition):

Screenshot from 2022-04-21 12-49-33

src/types.ts
```diff
@@ -126,6 +126,8 @@ export interface AdxTableSchema {
 export interface AdxColumnSchema {
   Name: string;
   CslType: string;
+  Type: string;
```
sunker, line 129: Rather than setting many of these values to an empty string, can't we make this optional?

sunker (approved):
Looks good! 

I guess this limitation is in monaco and not in kusto-monaco? Maybe it's worth registering a feature request?

andresmgot:
> I guess this limitation is in monaco and not in kusto-monaco? Maybe it's worth registering a feature request?

no, it was a kusto-monaco thing (because of how the functions are defined there) buut I found the culprit. The definition was not liking my default value and that caused that error. Setting `CslDefaultValue: '""'` did the trick, now it does not complain with no arguments :).

PR #364, opened 2022-04-22
Finish new editor set up

### LINKED ISSUE #224: Use @grafana/UI CodeEditor

andresmgot:
Closes #224 

Two last changes here for the new query editor:

 - Added some logic to reload the schema when the database changes. This allows to suggest new tables / columns in the editor.
 - I removed the feature flag, since this is now ready to be used ... but only works for Grafana >= 8.5. I replaced the feature flag with a version gate. Only people running the latest versions of Grafana will get the new editor.

### DIFF
--- a/src/components/RawQueryEditor.tsx
+++ b/src/components/RawQueryEditor.tsx
@@ -5,7 +5,8 @@ import { CodeEditor, Icon, Monaco, MonacoEditor, useStyles2 } from '@grafana/ui'
 import { QueryEditorResultFormat, selectResultFormat } from 'components/QueryEditorResultFormat';
 import { AdxDataSource } from 'datasource';
 import { KustoMonacoEditor } from 'monaco/KustoMonacoEditor';
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
+import { gte, valid } from 'semver';
 import { selectors } from 'test/selectors';
 import { AdxDataSourceOptions, AdxSchema, KustoQuery } from 'types';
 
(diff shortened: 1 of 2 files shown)


src/components/RawQueryEditor.tsx
```diff
@@ -73,23 +86,27 @@ export const RawQueryEditor: React.FC<RawQueryEditorProps> = (props) => {
...
-        }
+        setWorker(worker);
       });
   };
 
+  useEffect(() => {
+    if (worker && schema) {
+      // Populate Database schema with macros
+      Object.keys(schema.Databases).forEach((db) =>
+        Object.assign(schema.Databases[db].Functions, getFunctions(templateSrv.getVariables()))
```
sunker, line 97: Put result in `templateSrv.getVariables()` in a variable so we don't have to call it multiple times?

sunker (commented):
Nice! I like how this version based check avoids a breaking change in the plugin. 

The kusto editor in azure monitor logs seems to check whether the monaco loading promise is resolved before starting to iterate over the languages. I guess there must be a reason for that?

iwysiu (approved):
Nice! I hadn't seen version mocking before.

andresmgot:
> The kusto editor in azure monitor logs seems to check whether the monaco loading promise is resolved before starting to iterate over the languages. I guess there must be a reason for that?

I am not really sure what you mean. If you are referring to these promises:

https://github.com/grafana/grafana/blob/main/public/app/plugins/datasource/grafana-azure-monitor-datasource/components/LogsQueryEditor/QueryField.tsx#L40

```ts
    const promises = [
      datasource.azureLogAnalyticsDatasource.getKustoSchema(query.azureLogAnalytics.resource),
      getPromise(),
    ] as const;
```

That is waiting for the schema to be returned from the backend and that the monaco editor has been mounted. We are doing the same here:

https://github.com/grafana/azure-data-explorer-datasource/pull/364/files#diff-41b7d37eb0dcbd41e5fb3353f5573a2556be30861a97dffe943991fbcc9fffd1R94

```ts
    if (worker && schema) {
      // Populate Database schema with macros
      Object.keys(schema.Databases).forEach((db) => ...
    }
```

`worker` will only be set when the editor is mounted and the language is loaded and schema is only set when retrieved from the backend.

Are you referring to a different thing?

sunker:
> Are you referring to a different thing?

No I'm referring to the `getPromise` that resolves when the monaco editor. Okay good then just wanted to confirm since now it's done in a slightly different way in ADX and AM logs. Thanks!

PR #365, opened 2022-04-22
Add e2e tests for the new code editor

andresmgot:
The goal of this PR is to upgrade the Grafana version that the e2e tests are run against so we can test that the new code editor is properly loaded and that suggestions work as expected.


asimpson (approved):
:rocket: LGTM, nice work!!

Issue #353, opened 2022-04-12
Query Parser has difficulty parsing columns that contain json or arrays of json


sarahzinger:
**What happened**:
We are showing users invalid Where filters

**What you expected to happen**:
I was expecting all dropdown options in the query editor lead to valid syntax

**How to reproduce it (as minimally and precisely as possible)**:
- have a column that contains json (in our internal test db of DestinyActivities look at the Teams column). Here is an example:
```
| id | Teams |

| --------------- | --------------- |

| 123 | {"18":{"Standing":"Victory","Score":5,"TeamID":0},"19":{"TeamID":0,"Standing":"Defeat","Score":3}} |
```
Our Query builder tries to parse this json with Where filters like so:
![Screen Shot 2022-04-11 at 8 57 19 PM](https://user-images.githubusercontent.com/6620164/162857267-dfaaf790-24e0-4f10-9e84-6f8b6eedfdcf.png)
- however if you try any of these filters, you'll hit a syntax error. You'll see that the KQL query we send off is:
```
DestinyActivities
| where $__timeFilter(TimeGenerated)
| where Teams.18.Score == 0
| order by TimeGenerated asc
```
when I think it should actually be something like this
```
DestinyActivities
| where $__timeFilter(TimeGenerated)
| where todynamic(Teams)["18"]["Score"] == 0
| order by TimeGenerated asc
```

You can see the same thing happen for columns with arrays of json:
- have a column that contains an array of json (in our internal test data at grafana look at the Players column in DestinyActivities) which is formated like so:
```
| id | Players |

| --------------- | --------------- |

| 123 | [{"Values":{"timePlayedSeconds":97,"completed":0,"completionReason":255,"efficiency":0},"Score":0}] |
```
- in our query builder dropdown you'll see in a Where filter that we have some odd options available to choose from:
![Screen Shot 2022-04-11 at 8 01 20 PM](https://user-images.githubusercontent.com/6620164/162852201-bca6b35d-7e40-4865-9d90-71e6eb8524ab.png)
- if you try to make a query with any of these options, we get back a syntax error

A current workaround for this would be to let users edit the KQL query directly.

I think it's valid to say that our Query Builder can not handle every possible data format (what if you want to query an array of arrays of arrays of arrays of json of json of json, forever and ever, how would you even render that in a Where filter), I do think it's a bit odd to show users options that will always result in a syntax error as they are not valid column types. I wonder is it possible for us to:
[...]

PR #367, opened 2022-04-27
Filter dynamic columns from query builder

aangelisc:
This PR tackles #353 by not displaying columns with type `dynamic` as options in the Where/Aggregate/Group by clauses to prevent malformed queries being generated. 

Further discussion is then needed on how we handle these columns.

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,6 +294,16 @@ MyLogs
 | project Timestamp, Text=Message , Tags="tag1,tag2"
 ```
 
+## Query Builder - Data Types
+
+The query builder provides an easy to use interface to query Azure Data Explorer. However, there are limitations on the supported data types that a column can possess. Currently, if a column is typed as `dynamic` it is filtered as an option for the following operations: `Where`, `Aggregate`, `Group by`. The reason for this is that columns of type `dynamic` can potentially contain values that have any of the primitive data types, but also arrays (where the array can then have values of any type) and JSON objects. The query builder does not currently support querying values that are either arrays or JSON objects.
+
+See the below documentation for further details on how to handle dynamic columns appropriately via the KQL editor.
+
+[Kusto Data Types](https://docs.microsoft.com/en-us/azure/data-explorer/kusto/query/scalar-data-types/) - Documentation on data types supported by Kusto.
+
+[Dynamic Data Type](https://docs.microsoft.com/en-us/azure/data-explorer/kusto/query/scalar-data-types/dynamic) - Detailed documentation on the dynamic data type.
+
 ## CHANGELOG
 
 See the [Changelog](https://github.com/grafana/azure-data-explorer-datasource/blob/master/CHANGELOG.md).
--- a/README.md
+++ b/README.md
@@ -296,7 +296,7 @@ MyLogs
 
 ## Query Builder - Data Types
 
-The query builder provides an easy to use interface to query Azure Data Explorer. However, there are limitations on the supported data types that a column can possess. Currently, if a column is typed as `dynamic` it is filtered as an option for the following operations: `Where`, `Aggregate`, `Group by`. The reason for this is that columns of type `dynamic` can potentially contain values that have any of the primitive data types, but also arrays (where the array can then have values of any type) and JSON objects. The query builder does not currently support querying values that are either arrays or JSON objects.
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -294,6 +294,16 @@ MyLogs
 | project Timestamp, Text=Message , Tags="tag1,tag2"
 ```
 
+## Query Builder - Data Types
+
+The query builder provides an easy to use interface to query Azure Data Explorer. However, there are limitations on the supported data types that a column can possess. Currently, if a column is typed as `dynamic` it is filtered as an option for the following operations: `Where`, `Aggregate`, `Group by`. The reason for this is that columns of type `dynamic` can potentially contain values that have any of the primitive data types, but also arrays (where the array can then have values of any type) and JSON objects. The query builder does not currently support querying values that are either arrays or JSON objects.
```
andresmgot, line 299:
```suggestion
The query builder provides an easy to use interface to query Azure Data Explorer. However, there are limitations on the supported data types that a column can possess. Currently, if a column is typed as `dynamic` it is not included as an option for the following operations: `Where`, `Aggregate`, `Group by`. The reason for this is that columns of type `dynamic` can potentially contain values that have any of the primitive data types, but also arrays (where the array can then have values of any type) and JSON objects. The query builder does not currently support querying values that are either arrays or JSON objects.
```

src/components/VisualQueryEditor.tsx
```diff
@@ -95,6 +95,9 @@ export const VisualQueryEditor: React.FC<Props> = (props) => {
   const columns = useColumnOptions(tableSchema.value);
   const groupable = useGroupableColumns(columns);
 
+  const columnTooltip =
+    "Some columns may not be visible for selection. This is due to the type of the column. The visual query editor does not currently support columns of type 'dynamic'";
```
andresmgot, line 99:
```suggestion
    "Some columns may not be visible for selection. The visual query editor does not currently support columns of type 'dynamic'";
```

(the shorter, the better for tooltips)

src/editor/components/QueryEditorSection.tsx
```diff
@@ -3,12 +3,13 @@ import { InlineFormLabel } from '@grafana/ui';
...
 export interface QueryEditorSectionProps {
   label: string;
+  tooltip?: string;
 }
 
 export const QueryEditorSection: React.FC<PropsWithChildren<QueryEditorSectionProps>> = (props) => {
   return (
     <div className="gf-form">
-      <InlineFormLabel className="query-keyword" width={12}>
+      <InlineFormLabel className="query-keyword" width={12} tooltip={props.tooltip ?? ''}>
```
andresmgot, line 12: not sure if the component works that way but there is the risk that this renders a tooltip icon but with no content. Better to leave it as `undefined`:

```suggestion
      <InlineFormLabel className="query-keyword" width={12} tooltip={props.tooltip}>
```

src/schema/AdxSchemaResolver.ts
```diff
@@ -80,14 +80,17 @@ export class AdxSchemaResolver {
       );
 
       return schema.OrderedColumns.reduce((columns: AdxColumnSchema[], column) => {
-        const schemaForDynamicColumn = schemaByColumn[column.Name];
+        if (column.CslType !== 'dynamic') {
```
andresmgot, line 83: one of our goals for this plugin is to increase the code coverage. Can you add a new test file to test (at least) this new functionality?

andresmgot (commented):
Thanks @aangelisc! I have minor suggestions.

src/editor/components/QueryEditorSection.tsx
aangelisc: Makes sense!

src/schema/AdxSchemaResolver.ts
aangelisc: I will work on getting some tests in for this (and potentially expanding code coverage in a future PR).

sarahzinger (approved):
Seems like a great first step! Let's add this to the changelog, also would we say this is a breaking change or not really? I think it would be nice to add a test to ensure that if for some reason you did have a funky broken saved query that the page still loads even if it can't render your selection.

andresmgot (approved):
Thanks for the changes! Please don't forget to add some tests

andresmgot:
> Let's add this to the changelog, 

That'd be nice, that way whenever we release we don't need to remember about this

> also would we say this is a breaking change or not really?

I don't think so? Even if you were using a dynamic type there and working it wouldn't break until you open the editor and you select that value but still something nice to notice in the changelog.

andresmgot (approved):
LGTM! Just one minor comment for the test.

PR #368, opened 2022-04-27
Avoid crash when loading legacy kusto editor

andresmgot:
Ref: #325

There is already a new code editor for the plugin that avoids the issue of #325 but since people running Grafana <8.5 won't get the fix, this PR helps a bit with the UX. Rather than crashing and not rendering the editor we can catch the exception and render the editor with no Kusto functionality. I have also added a message to urge users to upgrade to newer versions:

Screenshot from 2022-04-28 13-22-47

Note that this only helps with the flow Grafana editor -> ADX editor. The other way around still hard-crash.

### DIFF
--- a/src/monaco/kusto_code_editor.ts
+++ b/src/monaco/kusto_code_editor.ts
@@ -22,6 +22,7 @@ export default class KustoCodeEditor {
   ) {}
 
   initMonaco(content: string) {
+    let error: Error | undefined;
     const themeName = this.config.bootData.user.lightTheme ? 'grafana-light' : 'vs-dark';
 
     monaco.editor.defineTheme('grafana-light', {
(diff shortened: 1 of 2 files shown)


src/monaco/kusto_code_editor.ts
```diff
@@ -50,11 +50,18 @@ export default class KustoCodeEditor {
...
-    });
+    try {
+      monaco.languages['kusto'].kustoDefaults.setLanguageSettings({
+        includeControlCommands: true,
+        newlineAfterPipe: true,
+        useIntellisenseV2: false,
+      });
+    } catch (e) {
+      console.error(
+        'Unable to load Kusto language. Try refreshing the page or upgrade to the latest version of Grafana and the ADX plugin.',
```
sarahzinger, line 61: maybe we can specify which version of grafana fixes the issue

sarahzinger (approved):
Seems like a good idea to me! Let's also add this one to the changelog

sunker:
Can we display an alert box underneath the code editor instead? Then the message will be picked up by more users.

andresmgot:
_added the message as an alert, see the description at the top_

sunker (approved):
Nice!

Issue #327, opened 2022-03-04
Increase unit-test coverage


andresmgot:
**What happened**:

We want at least a 50% coverage for all our plugins (both frontend and backend). At the moment, the coverage for ADX frontend code is 29.79%:

```
▶ yarn run jest --coverage
------------------------------------------|---------|----------|---------|---------|--------------------------------------------------------
File                                      | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s                                      
------------------------------------------|---------|----------|---------|---------|--------------------------------------------------------
All files                                 |   29.79 |    27.52 |   22.87 |   30.07 |                                                        
```

This can be highly affected by https://github.com/grafana/azure-data-explorer-datasource/issues/288

Note: the backend coverage is 57.1% so that is fine enough.

vickyyyyyyy:
note: we should probably do this during/after our larger piece of work of refactoring ADX

andresmgot:
After #369 the coverage is now ~60% so we can consider this task done.

PR #369, opened 2022-04-27
Increase test coverage

andresmgot:
Ref: #327 

I started with some folders with low coverage. This PR:

 - Removes common/annotationsFromFrame.ts since it was not being used
 - Adds a test for common/responseHelpers.ts
 - Moves  src/QueryEditor.tsx to the components folder and add tests for it


src/common/responseHelpers.test.ts
```diff
@@ -0,0 +1,23 @@
+import { DataFrame, FieldType } from '@grafana/data';
+import { firstStringFieldToMetricFindValue } from './responseHelpers';
+
+describe('firstStringFieldToMetricFindValue', () => {
+  it('should find a string field value', () => {
+    const frame: DataFrame = {
```
aangelisc, line 6: May be worth mocking the `DataFrame` type as I imagine this will be used elsewhere. Also helps with consistency across plugins.

aangelisc (approved):
LGTM as is - could be worth mocking the DataFrame as mentioned but can be done elsewhere😊

PR #372, opened 2022-04-29
MSAL for On-Behalf-Of token requests

kostrse:
Direct implementation of token HTTP request was replaced by implementation from MSAL for Go.

This is consistent with client secret token requests in Azure SDK ([here](https://github.com/Azure/azure-sdk-for-go/blob/a516802508c2239b7782b8c3fd26b6d582357e21/sdk/azidentity/client_secret_credential.go#L48) and [here](https://github.com/Azure/azure-sdk-for-go/blob/a516802508c2239b7782b8c3fd26b6d582357e21/sdk/azidentity/client_secret_credential.go#L65))

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/grafana/azure-data-explorer-datasource
 go 1.17
 
 require (
+	github.com/AzureAD/microsoft-authentication-library-for-go v0.4.0
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v0.13.2
 	github.com/google/go-cmp v0.5.7
 	github.com/grafana/grafana-azure-sdk-go v1.1.0
@@ -17,7 +18,6 @@ require (
 require (
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v0.22.0 // indirect
 	github.com/Azure/azure-sdk-for-go/sdk/internal v0.9.1 // indirect
-	github.com/AzureAD/microsoft-authentication-library-for-go v0.4.0 // indirect
 	github.com/apache/arrow/go/arrow v0.0.0-20211112161151-bc219186db40 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/cespare/xxhash/v2 v2.1.2 // indirect
--- a/pkg/azuredx/azureauth/aad_client.go
+++ b/pkg/azuredx/azureauth/aad_client.go
@@ -1,0 +1,69 @@
+package azureauth
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"regexp"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/confidential"
+	"github.com/grafana/grafana-azure-sdk-go/azcredentials"
+	"github.com/grafana/grafana-azure-sdk-go/azsettings"
+)
+
+// Abstraction over confidential.Client from MSAL for Go
+type aadClient interface {
+	AcquireTokenOnBehalfOf(ctx context.Context, userAssertion string, scopes []string) (confidential.AuthResult, error)
+}
+
+func newAADClient(credentials *azcredentials.AzureClientSecretCredentials, httpClient *http.Client) (aadClient, error) {
+	authority, err := resolveAuthorityForCloud(credentials.AzureCloud)
+	if err != nil {
+		return nil, fmt.Errorf("invalid Azure credentials: %w", err)
+	}
+
+	if !validTenantId(credentials.TenantId) {
+		return nil, errors.New("invalid tenantId")
+	}
+
+	clientCredential, err := confidential.NewCredFromSecret(credentials.ClientSecret)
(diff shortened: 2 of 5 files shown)


pkg/azuredx/azureauth/auth.go
```diff
@@ -147,71 +141,10 @@ func (c *ServiceCredentialsImpl) queryDataOnBehalfOf(ctx context.Context, req *b
...
+func (c *ServiceCredentialsImpl) onBehalfOf(ctx context.Context, idToken string) (onBehalfOfToken string, expire time.Time, err error) {
+	result, err := c.aadClient.AcquireTokenOnBehalfOf(ctx, idToken, c.scopes)
 	if err != nil {
-		return "", time.Time{}, fmt.Errorf("on-behalf-of grant POST <%q>: %w", tokenURL, err)
-	}
-	body, err := io.ReadAll(resp.Body)
-	if err != nil {
-		return "", time.Time{}, fmt.Errorf("on-behalf-of grant POST <%q> response: %w", tokenURL, err)
-	}
-	onBehalfOfLatencySeconds.WithLabelValues(strconv.Itoa(resp.StatusCode)).Observe(float64(time.Since(reqStart)) / float64(time.Second))
```
andresmgot, line 182: I am not sure if this was intended but you are removing the metric here. I am not sure if this metric is useful though but if we want to remove it, we should also remove the variable `onBehalfOfLatencySeconds`. If we remove it we should also make a note in the changelog. What do you think?

andresmgot (commented):
Code looks good to me! (I tried and is also working for my setup). I just have a question about some metric collection that was going on there.

pkg/azuredx/azureauth/auth.go
kostrse: I removed it intentionally because a single metric doesn't bring good picture and I didn't understand how it could be useful.
We could add metrics on the  Grafana Azure SDK layer, so they could be transparent to datasource and cover different scenarios (different auth types, cache hits etc.).

Removed references to the metric and updated the CHANGELOG.

Added task to Grafana Azure SDK: https://github.com/grafana/grafana-azure-sdk-go/issues/7

pkg/azuredx/azureauth/aad_client.go
```diff
@@ -0,0 +1,69 @@
...
+
+	client, err := confidential.New(credentials.ClientId, clientCredential, authorityOpts, httpClientOpts)
+	if err != nil {
+		return nil, err
+	}
+
+	return client, nil
+}
+
+func resolveAuthorityForCloud(cloudName string) (azidentity.AuthorityHost, error) {
```
sunker, line 48: We already have this func in the SDK, no? Can we use it instead?

pkg/azuredx/azureauth/aad_client.go
```diff
@@ -0,0 +1,69 @@
...
+	case azsettings.AzureUSGovernment:
+		return azidentity.AzureGovernment, nil
+	default:
+		err := fmt.Errorf("the Azure cloud '%s' not supported", cloudName)
+		return "", err
+	}
+}
+
+func validTenantId(tenantId string) bool {
+	match, err := regexp.MatchString("^[0-9a-zA-Z-.]+$", tenantId)
```
sunker, line 64: Feels like this could be moved to the SDK too?

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/grafana/azure-data-explorer-datasource
 go 1.17
 
 require (
+	github.com/AzureAD/microsoft-authentication-library-for-go v0.4.0
```
sunker, line 6: Seems like this lib is in production preview. Guess it's fine then? Is there a reason this is not part of the azure-sdk-go?

kostrse: `azure-sdk-go` does actually use it for client secret scenario, same version of MSAL library (indirect dependency changes to direct dependency).
They had their own implementation in past but replaced it by MSAL: [here](https://github.com/Azure/azure-sdk-for-go/blob/a516802508c2239b7782b8c3fd26b6d582357e21/sdk/azidentity/client_secret_credential.go#L48) and [here](https://github.com/Azure/azure-sdk-for-go/blob/a516802508c2239b7782b8c3fd26b6d582357e21/sdk/azidentity/client_secret_credential.go#L65)

Usage of MSAL in my `aad_client.go` is very similar to how they use it in their `client_secret_credential.go`.
Azure SDK doesn't support OBO scenario, so I decided to refence MSAL directly, similar how they do it for client secret.

pkg/azuredx/azureauth/aad_client.go
kostrse: The end goal is to move everything from here to the SDK (`AzureTokenProvider`), so I don't want to expose this API in SDK as it won't be needed in future.
This is only incremental change. Here I just want to get read of home made implementation of token client to have `auth.go` cleaner for further refactoring.

sunker: I'm fine with that!

go.mod
sunker: Okay, thanks for making this clear!

pkg/azuredx/azureauth/auth.go
andresmgot: okay, sounds good

andresmgot (approved):
LGTM, thanks!

PR #377, opened 2022-05-12
ConfigEditor: make tooltip clickable and fix text wrapping on button

asimpson:


andresmgot (commented):
once you are back, you will need to merge main here before :)

PR #378, opened 2022-05-13
Read AzureSettings from Grafana host via SDK

### LINKED ISSUE #357: Azure settings should be received from the Grafana host
Currently Azure settings assume public cloud.
This is not critical for App Registration (client secret) authentication because user explicitly select a cloud, but it will be important for Managed Identity authentication because in that case the cloud should be the same as where Grafana hosted.

kostrse:
Use Grafana Azure SDK ([github.com/grafana/grafana-azure-sdk-go/azsettings/env.go](https://github.com/grafana/grafana-azure-sdk-go/blob/e8d4106df2fc474a8b7f55dd3b82f92868d9ed27/azsettings/env.go#L20)) to read `azsettings.AzureSettings` from environment variables passed by the Grafana host.

Fixes #357

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,20 +3,19 @@ module github.com/grafana/azure-data-explorer-datasource
 go 1.17
 
 require (
-	github.com/AzureAD/microsoft-authentication-library-for-go v0.4.0
+	github.com/Azure/azure-sdk-for-go/sdk/azcore v0.22.0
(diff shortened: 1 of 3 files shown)


kostrse:
cc @sunker

kevinwcyu (approved):
LGTM, I'll let @sunker have a look too.

sunker (approved):
Looks good, thank you @kostrse!

Issue #359, opened 2022-04-14
On-Behalf-Of feature should be under a feature flag


kostrse:
The On-Behalf-Of feature has a few unresolved edge cases:
* It won't work when AAD environment of Grafana (and thus of user's ID token) is different than Azure environment of the datasource (ADX cluster).
* It won't work when configured app client doesn't have consent to access the service because there's no ability for interactive consent flow.
* It cannot work when Grafana uses non-AAD authenticaton method but UI doesn't reflect it.

Considering this I would put On-Behalf-Of under a feature flag until a more complete solution available.

This way users who well understand the shortcomings would be able to explicitly enable it for their particular scenario but it wouldn't confuse other users until it's polished.

sunker:
It's already stated in the [docs](https://github.com/grafana/azure-data-explorer-datasource/blob/main/doc/on-behalf-of.md) and in the ds configuration ui that this feature is in beta so I'm not sure that's necessary. 

Since the OBO feature in ADX is under a beta flag, we're allowed to make changes to how it behaves and how it's being used. 

While adding support for the edge cases you've described, it might be possible to address it in a few different ways:
1) in a custom middleware in the ADX plugin
2) a part of the Azure middleware in the Azure SK
3) a generic user auth feature in Grafana (discussed [here](https://github.com/grafana/grafana/discussions/39701)) 

I'm curious to hear your thoughts on this @kostrse.

kostrse:
> It's already stated in the [docs](https://github.com/grafana/azure-data-explorer-datasource/blob/main/doc/on-behalf-of.md) and in the ds configuration ui that this feature is in beta so I'm not sure that's necessary.

@sunker could you please give details how does "beta flag" work? Is it something that can be switched on/off in Grafana configuration?

For Azure Grafana service we need this OBO toggle to not be available for users, since it doesn't work there.
It should be something like a feature toggle which we could set in Grafana config.

yaelleC:
After discussion with the team, it seems only ASML is using the feature so the default state will be disabled.

yaelleC:
Update :

- Grafana PR for Feature flag creation : [#49476](https://github.com/grafana/grafana/pull/49476)
- PR for Frontend changes to hide toggle in ADX : #387

Unfortunately: 
[...]

PR #387, opened 2022-05-24
Config Editor : Placing OBO functionality under feature gate

yaelleC:
Fixes : #359 

**Done:**

- [x] Hiding OBO toggle when feature gate disabled
- [x] Disabling OBO when updating a datasource, even if the user enabled it on plugin config previously.


src/components/ConfigEditor/ConnectionConfig.tsx
```diff
@@ -164,33 +165,36 @@ const ConnectionConfig: React.FC<ConnectionConfigProps> = ({
...
-            value={jsonData.onBehalfOf}
-            onChange={(ev: React.ChangeEvent<HTMLInputElement>) => updateJsonData('onBehalfOf', ev.target.checked)}
-          />
-          <span>⚠️ This feature is in beta</span>
-        </HorizontalGroup>
-      </InlineField>
+      {config.featureToggles.adxOnBehalfOf && (
+        <InlineField
+          label="Use On-Behalf-Of"
+          htmlFor="adx-on-behalf-of"
```
yaelleC, line 171: Adding `htmlFor` explicitly as the `getChildId` used in `InlineField` will only look at first level children (i.e.`HorizontalGroup` here).

Having a correct `for` wiring helps with RTL and a11y 🚀

README.md
```diff
@@ -6,7 +6,8 @@
 
 ## Installation
 
-This plugin requires Grafana 7.1.0 or newer as of version 3.0.0. Plugin versions prior to 3.0.0 require Grafana 6.3.6.
+This plugin requires Grafana 8.0.0 or newer as of version 4.0.0 , it requires Grafana 7.1.0 or newer as of version 3.0.0.
```
yaelleC, line 9: Missing in current readme, but not related to this PR

pkg/azuredx/azureauth/auth.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/grafana/grafana-azure-sdk-go/azsettings"
 	"github.com/grafana/grafana-azure-sdk-go/aztokenprovider"
 	"github.com/grafana/grafana-plugin-sdk-go/backend"
+	"github.com/grafana/grafana/pkg/services/featuremgmt"
```
andresmgot, line 15: mmh, I am not sure you can do this, you will not be able to use inner packages from Grafana in plugins :thinking: Maybe you can ask in #grafana-backend or in #grafana-plugins-platform if it's possible to use backend feature flags in plugins? There may be some other way of doing it but we don't have any examples in the plugins we own.

yaelleC: Yes, I had the same feeling, I was looking into external plugins on Github with no luck either.

I've removed the backend code, 
As is the branch will prevent OBO feature to be used with newly created data source when feature gate is OFF, but it will not prevent the query being done using OBO if it was enabled when the data source was created. Two ways forwards it seems:

1) _Ideal if possible_: I wonder if we can pass the `onBehalfOf = false` from the frontend when executing the query?
2) _Less user friendly_: I add code in the config editor (maybe here: https://github.com/grafana/azure-data-explorer-datasource/blob/15af456286d443f6499c714965a9ea6166420e84/src/components/ConfigEditor/ConnectionConfig.tsx#L33) to check the feature flag value. Meaning users would have to open and save the data source for the change to be applied in the data source `jsonData`

src/components/ConfigEditor/ConnectionConfig.tsx
```diff
@@ -164,33 +165,36 @@ const ConnectionConfig: React.FC<ConnectionConfigProps> = ({
...
-        <HorizontalGroup>
-          <InlineSwitch
-            id="adx-on-behalf-of"
-            value={jsonData.onBehalfOf}
-            onChange={(ev: React.ChangeEvent<HTMLInputElement>) => updateJsonData('onBehalfOf', ev.target.checked)}
-          />
-          <span>⚠️ This feature is in beta</span>
-        </HorizontalGroup>
-      </InlineField>
+      {config.featureToggles.adxOnBehalfOf && (
```
andresmgot, line 168: I would also add a check in the first useEffect her (line 30) so in case an existing data source has the feature enabled but it's disabled by configuration it updates the value:

```ts
    if (jsonData.onBehalfOf && !config.featureToggles.adxOnBehalfOf) {
      updateJsonData('onBehalfOf', false);
    }
```

pkg/azuredx/azureauth/auth_test.go
```diff
@@ -87,6 +87,34 @@ func TestOnBehalfOf(t *testing.T) {
 	}
 }
 
+// TODO test feature flag disabled too -> should have same result
+func TestOnBehalfOfDisabled(t *testing.T) {
```
andresmgot, line 91: elaborated in https://github.com/grafana/azure-data-explorer-datasource/issues/359#issuecomment-1136964347 but I don't think we _need_ the backend changes.

yaelleC: I'll remove the TODO but will leave the test as currently it checks that if the feature is disabled in the data source it will not use OBO (i.e. use TokenProvider, not AADClient), we did not have a test for this 😉

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Change Log
 
+## [4.1.0]
```
yaelleC, line 3: a breaking change in a beta feature did not seem worth a major release to me, but also too big to warrant a patch one... WTDY?

andresmgot: I think this can qualify more as a fix than a new feature, so I would do 4.0.2. I agree with not releasing a new major.

BTW, remember to update the package.json :+1:

README.md
```diff
@@ -153,7 +154,17 @@ If the command succeeds you should get a result like this:
 
 6. Click the `Save & Test` button. After a few seconds once Grafana has successfully connected then choose the default database and save again.
 
-[on-behalf-of documentation](https://github.com/grafana/azure-data-explorer-datasource/blob/main/doc/on-behalf-of.md)
+changes to ### On-Behalf-Of (Beta)
```
andresmgot, line 157:
```suggestion
#### Configuring On-Behalf-Of (Beta) authentication
```

README.md
```diff
@@ -153,7 +154,17 @@ If the command succeeds you should get a result like this:
...
+⚠️ _This feature is in Beta and subject to breaking changes_
+
+Since 3.7.0, we provide OBO flow for Token Authorization. To be used, this feature needs to be enabled explicitely in Grafana's config:
+
+```ini
+# On-Behalf-Of functionality for the ADX data source
+adxOnBehalfOf = true
+```
```
andresmgot, line 165: I would move this info to the other doc to have everything there

pkg/azuredx/azureauth/auth_test.go
```diff
@@ -86,6 +86,32 @@ func TestOnBehalfOf(t *testing.T) {
...
+	c := &ServiceCredentialsImpl{
+		tokenProvider:  fakeTokenProvider,
+	}
+	c.OnBehalfOf = false
+
+	auth, err := c.QueryDataAuthorization(context.Background(), &req)
+
+	if err != nil {
+		t.Errorf("got error %q", err)
+	}
```
andresmgot, line 108: another, shorter, way of writing this: `require.NoError(t, err)`

`require` from `"github.com/stretchr/testify/require"`

pkg/azuredx/azureauth/auth_test.go
```diff
@@ -86,6 +86,32 @@ func TestOnBehalfOf(t *testing.T) {
...
+
+	if !fakeTokenProvider.TokenRequested {
+		t.Errorf("got %q, expected 'TokenRequested", auth)
+	}
```
andresmgot, line 112: same: `assert.Equal(t, fakeTokenProvider.TokenRequested, true)`

but for clarity with the previous set of tests, shouldn't you use a FakeAADClient and check that `TokenRequested` is false? (Therefore, `AcquireTokenOnBehalfOf` has not been called). 

What I would do is to extend the table test above, make `OnBehalfOf` a test input and check that the token has not been requested. Something like:

```
		4: {
			RequestUser:          &backend.User{Login: "alice"},
			RequestIDTokenHeader: "ID-TOKEN",
			OnBehalfOfEnabled: false,
			ShouldRequestToken:   false,
			ExpectedError:        "",
		},
```
...
```
	c.OnBehalfOf = g.OnBehalfOfEnabled
```

src/components/ConfigEditor/ConfigEditor.test.tsx
```diff
@@ -44,4 +44,47 @@ describe('ConfigEditor', () => {
...
+    const originalConfigValue = grafanaRuntime.config.featureToggles.adxOnBehalfOf;
+
+    grafanaRuntime.config.featureToggles.adxOnBehalfOf = true;
+
+    render(<ConfigEditor {...mockConfigEditorProps()} />);
+
+    await waitFor(() => expect(screen.getByLabelText('Use On-Behalf-Of')).toBeInTheDocument());
+
+    // reset config to not impact future tests
+    grafanaRuntime.config.featureToggles.adxOnBehalfOf = originalConfigValue;
```
andresmgot, line 58: this needs to be on a `beforeEach` hook: if the test fails, this line is not executed.

andresmgot (commented):
LGTM, mostly minor comments!

CHANGELOG.md
yaelleC: Thanks, yes I was waiting for the version to update the `package.json` .

Happy to go with `4.0.2` 👍

src/components/ConfigEditor/ConfigEditor.test.tsx
yaelleC: oops indeed! 🤭  Copy pasted without thinking.

doc/on-behalf-of.md
```diff
@@ -4,6 +4,13 @@
 
 ⚠️ Only compatible with Grafana 8.3.4 or above.
 
+Since 3.7.0, we provide OBO flow for Token Authorization. To be used, this feature needs to be enabled explicitely in Grafana's config:
+
+```ini
+# On-Behalf-Of functionality for the ADX data source
+adxOnBehalfOf = true
+```
```
andresmgot, line 12: I think we should mention that it's a feature flag, and the code to be:

```ini
[feature_toggles]
# On-Behalf-Of functionality for the ADX data source
adxOnBehalfOf = true
```

andresmgot (approved):
LGTM! Just one minor suggestion for the docs

sunker (commented):
Looks good! Just added a question.

src/components/ConfigEditor/ConnectionConfig.tsx
```diff
@@ -27,6 +28,9 @@ const ConnectionConfig: React.FC<ConnectionConfigProps> = ({
 
   // Set some default values
   useEffect(() => {
+    if (jsonData.onBehalfOf && !config.featureToggles.adxOnBehalfOf) {
+      updateJsonData('onBehalfOf', false);
```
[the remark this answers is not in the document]
andresmgot: sure, yes, I mean for people that don't have access to changing the configuration.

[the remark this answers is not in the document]
yaelleC: Yes, @sunker plus it felt weird having to enable the feature flag just to set the toggle to OFF and then disable the feature flag again.  

Clearly having a backend check was the preferred way but we went with what we felt was most consistent possible: if feature flag is disabled then any data source should have OBO disabled too , we can't do it for the existing ones, but we enforce it for the updated ones.

Are you happy for me to merge?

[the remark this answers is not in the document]
sunker: Yes totally fine! Just wanted to make sure you were aware of the consequences, but had I read your previous discussions I wouldn't have to ask. :)

PR #389, opened 2022-05-26
Add basic objects as options in the query builder

andresmgot:
I have split #353 in different subtasks, in this PR, I am adding support for dynamic columns that doesn't contain any array within them.

I noticed that there is some code already that generates a query that inspect tables in order to get dynamic columns information so we already can get the structure of a dynamic column. If a column contains an array, it contains the identifier ``indexer``, rather than a name so we can filter those out for the moment.

I have also put this behind a feature flag so it doesn't interfere with release until this is released.

Screenshot from 2022-05-26 15-59-03

### DIFF
--- a/src/datasource.ts
+++ b/src/datasource.ts
@@ -381,7 +381,9 @@ const recordSchema = (columnName: string, schema: any, result: AdxColumnSchema[]
   }
 
   for (const name of Object.keys(schema)) {
-    const key = `${columnName}.${name}`;
+    // Using > as a key separator since it's not a valid character for an identifier
+    // https://docs.microsoft.com/en-us/azure/data-explorer/kusto/query/schema-entities/entity-names#identifier-naming-rules
+    const key = `${columnName}>${name}`;
 
     if (typeof schema[name] === 'string') {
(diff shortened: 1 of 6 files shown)


src/datasource.ts
```diff
@@ -381,7 +381,9 @@ const recordSchema = (columnName: string, schema: any, result: AdxColumnSchema[]
   }
 
   for (const name of Object.keys(schema)) {
-    const key = `${columnName}.${name}`;
```
andresmgot, line 384: A columns can contain `.` in their names so in order to be able to identify dynamic columns just with the name, I have replaced it with `>`

aangelisc (approved):
LGTM, great start on this! There is one caveat to be aware of (which I'd say is an edge case but it may not be) - when we retrieve the dynamic schema we limit the number of rows to 50,000 which means there could potentially be properties that are unaccounted for. I wonder if we should potentially notify the user of this?

andresmgot:
> I wonder if we should potentially notify the user of this?

Probably something that we can add to the documentation :+1:

PR #390, opened 2022-06-01
Add support for dynamic objects in Aggregate and Group by fields

andresmgot:
I noticed that `dynamic` values cannot be used for the summarize function (used for the Aggregate and Group by fields). For aggregating, I am converting the value to a `long` since its goal is to be used with a mathematical function. For grouping I am converting the value to a string. Note that the goal of this is not to support all the possible use cases but to at least give valid syntax:

Aggregate example:
Screenshot from 2022-06-01 11-18-41

Group by example:
Screenshot from 2022-06-01 11-19-07

I tried to do some unit test for this but it seems that the current components are not working with the react testing library (clicking on the "+" button didn't render the selector). I could do some e2e test for covering this functionality but it doesn't seem worth it. Let me know what you think.

### DIFF
--- a/src/components/VisualQueryEditor.test.tsx
+++ b/src/components/VisualQueryEditor.test.tsx
@@ -45,7 +45,7 @@ const schema = {
 };
 
 describe('VisualQueryEditor', () => {
-  xit('should render the VisualQueryEditor', async () => {
+  it('should render the VisualQueryEditor', async () => {
     render(<VisualQueryEditor {...defaultProps} />);
     await waitFor(() => screen.getByText('Table schema loaded successfully but without any columns'));
   });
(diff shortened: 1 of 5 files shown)


src/components/VisualQueryEditor.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+      },
+      ExternalTables: {},
+      Functions: {},
+      MaterializedViews: {},
+    },
+  },
+};
+
+describe('VisualQueryEditor', () => {
+  xit('should render the VisualQueryEditor', async () => {
```
aangelisc, line 48: Why is this test marked to be skipped?

aangelisc (approved):
LGTM!  Just a minor question.

I'm happy with leaving the e2e test out as I agree it will be excess work for little gain. I think opening a separate ticket to investigate why the unit tests don't work as expected is probably a better course of action. Once that's solved we can unit test accordingly 👍

PR #392, opened 2022-06-01
Avoid unnecessary todynamic

andresmgot:
There is really no need to cast a dynamic value when you get it because the result is already a `dynamic` type. From:

https://docs.microsoft.com/en-us/azure/data-explorer/kusto/query/scalar-data-types/dynamic#dynamic-object-accessors

> Accessing a sub-object of a dynamic value yields another dynamic value, even if the sub-object has a different underlying type

Since there is no need to use `todynamic`, I can remove the logic I added #389 to use `>` in `dynamic` types and simply return a valid accessor when parsing the dynamic schema.

Note for the reviewer: There are a bunch of test changes in this PR but the functional changes are few, mostly removing extra handling:

[src/datasource.ts](https://github.com/grafana/azure-data-explorer-datasource/pull/392/files#diff-804a5756e358074cb0ba27b1932cca77d39b90694fc1d4d122f46348c8653259)
[src/schema/mapper.ts](https://github.com/grafana/azure-data-explorer-datasource/pull/392/files#diff-b5407913a74a840627ce3902427cb64abc474c938069ef2dfffd08123fafd448)
[src/components/VisualQueryEditor.tsx](https://github.com/grafana/azure-data-explorer-datasource/pull/392/files#diff-e4544781782f9bfb36cc56593b96635368f89bdc69811bcb009f9979bbe7a6d6)

### DIFF
--- a/src/components/QueryEditor.test.tsx
+++ b/src/components/QueryEditor.test.tsx
@@ -25,7 +25,7 @@ jest.mock('@grafana/runtime', () => {
 const defaultProps = {
   onChange: jest.fn(),
   onRunQuery: jest.fn(),
-  datasource: mockDatasource,
+  datasource: mockDatasource(),
   query: mockQuery,
 };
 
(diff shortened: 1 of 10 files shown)


aangelisc (approved):
LGTM 😊

PR #398, opened 2022-06-03
Support for dynamic nested arrays

### LINKED ISSUE #353: Query Parser has difficulty parsing columns that contain json or arrays of json

andresmgot:
This is the last PR to fully add support for dynamic objects in the query builder :tada: 

The main goal of this PR is to add support for dynamic objects containing nested arrays.

The list of changes are:

 - I have replaced the apperances of `mv-apply` with `mv-expand`. Both functions are similar and just using `mv-expand` allows me to define nested arrays. For example: 

```kusto
| mv-apply e = Players on (where e.Score  == 1)
```

is the same than:

```kusto
| mv-expand e = Players
| where e.Score == 1
```

 - I noticed that `mv-expand` (and `mv-apply`) generated a new column that was returned to the user. Since this is supposed to be a temporary calculation, I have added the logic to remove those (with [`project-away`](https://docs.microsoft.com/en-us/azure/data-explorer/kusto/query/projectawayoperator))
 - I have refactored the logic that calls `mv-expand` so it's reused both when using `where` and `summarize`
 - Added support for nested arrays. This is done calling recursively to `addExpanPartsdIfNeeded` and thanks to the above.
 - Removed the tooltips that warned about missing dynamic types and the feature flag since this should be completed :tada:

Screenshot from 2022-06-03 17-31-32
 

Closes #353

### DIFF
--- a/src/KustoExpressionParser.test.ts
+++ b/src/KustoExpressionParser.test.ts
@@ -832,6 +832,34 @@ describe('KustoExpressionParser', () => {
       );
     });
 
+    it('should parse expression with summarize function in a nested array', () => {
+      const expression = createQueryExpression({
+        from: createProperty('StormEvents'),
+        reduce: createArray([
+          createReduceWithParameter('column["`indexer`"]["foo"]["`indexer`"]', 'percentile', [1, '2']),
+        ]),
+      });
+
+      const tableSchema: AdxColumnSchema[] = [
+        {
+          Name: 'column["`indexer`"]["foo"]["`indexer`"]',
+          CslType: 'string',
+        },
+        {
+          Name: 'StartTime',
+          CslType: 'datetime',
+        },
+      ];
+
+      expect(parser.toQuery(expression, tableSchema)).toEqual(
+        'StormEvents' +
+          '\n| where $__timeFilter(StartTime)' +
+          '\n| mv-expand array_1 = column' +
+          '\n| mv-expand array_2 = array_1["foo"]' +
+          `\n| summarize percentile(tostring(array_2), 1, '2')`
+      );
+    });
+
     it('should parse expression with timeshift', () => {
       const expression = createQueryExpression({
         from: createProperty('StormEvents'),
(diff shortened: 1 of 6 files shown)


aangelisc (approved):
Amazing work getting all of this in 👍

Issue #391, opened 2022-06-01
Revamp query builder


andresmgot:
**What happened**:

The current query builder is based on custom components created especifically for this plugin. The development there has been stale for quite long so it has become brittle and difficult to maintain.

New and existing plugins are migrating to the new package that https://github.com/grafana/grafana-experimental exposes. We should start migrating the editor here to use the components from there. This should be behind a feature flag (or a version flag) until the builder is moved to the @grafana/ui package or is in the minimum version supported.

**What you expected to happen**:

To have a standard UX and a query builder easy to maintain.

**How to reproduce it (as minimally and precisely as possible)**:

**Screenshots**

**Anything else we need to know?**:

**Environment**:

- Grafana version:
- Plugin version:
- OS Grafana is installed on:
- User OS & Browser:
- Others:

**TODO**:
 - [X] Query editor header (effort/small)
 - [x] Raw query editor (effort/medium)
 - [x] #461

PR #413, opened 2022-06-20
Fix template variable quoting

andresmgot:
The goal of this PR is to automatically quote variables when used as strings in the query builder.

The main difficulty here is that not all variables should be quoted, only when comparing strings it makes sense to quote them. This means that we cannot just modify the behavior of the template service to quote everything because we still want to maintain other variables un-quoted, based on the schema and the position.

We can do that in the component that handle filters for the query: `filter/QueryEditorFieldAndOperator.tsx`. Here, we have access to the target property (and its type) and the list of template variables.

Another change is that before we were not handling labels in the selector. The labels were just equal to the value. I need to modify that because I want to show the label unquoted, while the final value should contain the single quotes. This behavior is similar to the editor from @grafana/experimental code so it should be easier to migrate once we tackle #391.

Screenshot from 2022-06-20 15-56-03

Screenshot from 2022-06-20 15-55-51

One final change is that it's now allowed to write a custom value for the filter, in case a different format is needed:

Screenshot from 2022-06-20 15-59-10

### DIFF
--- a/src/KustoExpressionParser.ts
+++ b/src/KustoExpressionParser.ts
@@ -16,6 +16,7 @@ import {
   QueryEditorPropertyExpression,
 } from './editor/expressions';
 import { cloneDeep } from 'lodash';
+import { SelectableValue } from '@grafana/data';
 
 interface ParseContext {
   timeColumn?: string;
@@ -289,16 +290,17 @@ export class KustoExpressionParser {
       return `(${value.map((v) => this.formatValue(v, type)).join(', ')})`;
     }
 
-    if (this.isTemplateVariable(value)) {
-      return value;
+    const val = typeof value === 'object' ? value.value : value;
+    if (this.isTemplateVariable(val)) {
+      return val;
     }
 
     switch (type) {
       case QueryEditorPropertyType.Number:
       case QueryEditorPropertyType.Boolean:
-        return value;
+        return val;
       default:
-        return `'${value}'`;
+        return `'${val}'`;
     }
   }
 
@@ -342,13 +344,15 @@ export class KustoExpressionParser {
     return property;
   }
 
-  private isTemplateVariable(value: string): boolean {
+  private isTemplateVariable(value: string | SelectableValue<string>): boolean {
     if (!Array.isArray(this.templateSrv.getVariables())) {
       return false;
     }
 
+    const val = typeof value === 'string' ? value : value.value || '';
+
     return !!this.templateSrv.getVariables().find((variable: any) => {
-      return `$${variable?.id}` === value || value.startsWith(`\$\{${variable?.id}:`);
+      return val.includes(`$${variable?.id}`) || val.startsWith(`\$\{${variable?.id}:`);
     });
   }
 }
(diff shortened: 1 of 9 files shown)


src/editor/guards.ts
```diff
@@ -54,7 +55,21 @@ export const isDateTimeOperator = (
   return propertyType === QueryEditorPropertyType.DateTime && typeof operator?.value === 'string';
 };
 
-export const isMultiOperator = (operator?: QueryEditorOperator): operator is QueryEditorOperator<string[]> => {
+export const isStringArray = (value: any): value is string[] => {
```
aangelisc, line 58: This could be simplified slightly to be:

```
if (!value instanceof Array) return false;
if (value.every(item => typeof item !== 'string') return false;
return true;
```

src/KustoExpressionParser.ts
```diff
@@ -342,13 +344,15 @@ export class KustoExpressionParser {
...
+  private isTemplateVariable(value: string | SelectableValue<string>): boolean {
     if (!Array.isArray(this.templateSrv.getVariables())) {
       return false;
     }
 
+    const val = typeof value === 'string' ? value : value.value || '';
+
     return !!this.templateSrv.getVariables().find((variable: any) => {
-      return `$${variable?.id}` === value;
+      return val.includes(`$${variable?.id}`) || val.startsWith(`\$\{${variable?.id}:`);
```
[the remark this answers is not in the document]
yaelleC: If it is the expected new behaviour, maybe we can add a test case to show that we thought of it and are happy with the logic?

src/editor/components/filter/QueryEditorFieldAndOperator.tsx
```diff
@@ -121,9 +129,10 @@ export const QueryEditorFieldAndOperator: React.FC<Props> = (props: Props) => {
       <QueryEditorField
         value={value?.property}
         fields={fields}
-        templateVariableOptions={templateVariableOptions}
+        templateVariableOptions={props.templateVariableOptions}
         onChange={onFieldChanged}
         placeholder="Choose column..."
+        aria-label="choose column"
```
yaelleC, line 135: This `aria-label` will be read to screen readers users possibly twice (some screen readers will read out the placeholder too) without context (the `Where (filter)` does not have an aria-label) so not very accessible.

* From an a11y perspective, using an aria-label in the label and a labelled-by here would be the preferred option
* For testing we could use an `id` here

Alternatively, as an easy but less best-practicy way we could have the `aria-label` here more descriptive (e.g. `choose column for where filter` )?

andresmgot: the underlying AsyncSelect is tricky to update so I'll go with updating the aria label

src/editor/components/operators/QueryEditorSingleOperator.tsx
```diff
@@ -88,6 +86,7 @@ export class QueryEditorSingleOperator extends PureComponent<Props, State> {
         backspaceRemovesValue
         isClearable
         noOptionsMessage="No options found"
+        aria-label="select value"
```
yaelleC, line 89: idem a11y here: this message will not add much info for a screen reader?

yaelleC (approved):
LGTM

Issue #417, opened 2022-06-24
Query Builder: Wrong syntax when a dynamic column maps to a primitive scalar data type


andresmgot:
**What happened**:

This bug is similar to https://github.com/grafana/azure-data-explorer-datasource/issues/411. The problem is that the code that parses the schema, it assumes that a dynamic column will contain an object, so it execute `Object.keys` over it:

https://github.com/grafana/azure-data-explorer-datasource/blob/324c50328eb3622610a2c67168ebfd4fea753be2/src/datasource.ts#L384

This doesn't work if the dynamic column is a single primitive type (e.g. `double`) or an array of them (e.g. `[double, long`]).

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

The easiest way to reproduce it is to create the following test cases in `datasource.test.ts` (line 383):

```patch
        { schema: '["long","double"]', expected: { Name: `Teams`, CslType: 'double' } },
        { schema: '"long"', expected: { Name: `Teams`, CslType: 'long' } },
```

which fails at the moment:

```
 FAIL  src/datasource.test.ts
  ● AdxDataSource › when getting a dynamic schema › when there are multiple types returned › should return [object Object] type

    expect(received).toEqual(expected) // deep equality

    - Expected  - 1
    + Received  + 5

      Object {
        "Teams": Array [
          Object {
    +       "CslType": "long",
    +       "Name": "Teams[\"0\"]",
    +     },
    +     Object {
            "CslType": "double",
    -       "Name": "Teams",
    +       "Name": "Teams[\"1\"]",
          },
        ],
      }

      413 |           });
      414 |
    > 415 |           expect(await datasource.getDynamicSchema('foo', 'bar', ['col'])).toEqual({
          |                                                                            ^
      416 |             Teams: [t.expected],
      417 |           });
      418 |           if (t.warn) {

      at next (src/datasource.test.ts:415:76)
      at fulfilled (node_modules/tslib/tslib.js:114:62)

  ● AdxDataSource › when getting a dynamic schema › when there are multiple types returned › should return [object Object] type

    expect(received).toEqual(expected) // deep equality

    - Expected  -  2
    + Received  + 14

      Object {
        "Teams": Array [
          Object {
    -       "CslType": "long",
    -       "Name": "Teams",
    +       "CslType": "l",
    +       "Name": "Teams[\"0\"]",
    +     },
    +     Object {
    +       "CslType": "o",
[...]

PR #424, opened 2022-07-11
Query builder: Fix dynamic resolution for simple types

andresmgot:
Fixes https://github.com/grafana/azure-data-explorer-datasource/issues/417

The `dynamic` type can be a simple type, an object or an array of simple types or objects. We were only considering the option that the type was an object.

Additional notes:
 - I have defined a new type that should cover the above, avoiding using the type `any`
 - I have removed some unused tests (that didn't reference any real code)


aangelisc (approved):
LGTM 😊 Nice work!

Issue #403, opened 2022-06-09
Removing column in group by clause removes the last one and not the selected one


shahdevangp:
![removing group by clause](https://user-images.githubusercontent.com/46785787/172815481-392f2cb1-0165-4541-b82d-24633b1680aa.gif)

**What happened**: In Query Builder, when using multiple columns in Group By, on clicking the "-" button next to each column, then it always removes the last column in the group by clause. It doesnt remove the group by column for which "-" was pressed.

**What you expected to happen**: Remove the group by column for which "-" was pressed and not the last column.

**How to reproduce it (as minimally and precisely as possible)**: See the gif.

**Environment**:

- Grafana version: 8.5.5
- Data source Plugin version: dont know how to determine that
- OS Grafana is installed on: Grafana Cloud
- User OS & Browser: Windows 11 Microsoft Edge

sunker:
Work started in a mob sessions. See progress [here](https://github.com/grafana/azure-data-explorer-datasource/tree/mob/main-remove-column-bug). We did not get far, so will remove ourselves as assignees.

PR #425, opened 2022-07-11
Fix "Aggregate" and "Group by" removal logic

andresmgot:
Fixes https://github.com/grafana/azure-data-explorer-datasource/issues/403

This was a rendering issue, the right element was being removed from the array but the component didn't re-render with the new data.

Peek 2022-07-11 14-41

I am not adding tests because this code should go away once we tackle https://github.com/grafana/azure-data-explorer-datasource/issues/391


yaelleC (approved):
LGTM

PR #426, opened 2022-07-11
Fix: return configured database instead of the first one.

### LINKED ISSUE #419: Default database setting has no effect in the query editor
**What happened**:

After changing the default database for an ADX data source, new panels still use the first default database.

The default database to be use by default in the database selector in the query editor.

- create an ADX adata source with a default database
- change the default database, and click save and test
- create a nex dashboard and a new panel, and add a query to that data source
- observe that it still uses the first database.

**Screenshots**

**Environment**:

- Grafana version:
- Plugin version: 4.1.1
- OS Grafana is installed on:
- User OS & Browser:
- Others:

asimpson:
Previously we just returned the first configured database in the query editor database dropdown. This now compares the default DB to the list of DBs and returns the match if there is one.

image

Fixes #419

### DIFF
--- a/src/components/QueryEditor.tsx
+++ b/src/components/QueryEditor.tsx
@@ -140,6 +140,7 @@ const useSelectedDatabase = (
   query: KustoQuery,
   datasource: AdxDataSource
 ): string => {
+  const defaultDB = useAsync(() => datasource.getDefaultOrFirstDatabase(), [datasource]);
   const variables = datasource.getVariables();
 
   return useMemo(() => {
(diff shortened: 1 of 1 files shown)


src/components/QueryEditor.tsx
```diff
@@ -140,6 +140,7 @@ const useSelectedDatabase = (
   query: KustoQuery,
   datasource: AdxDataSource
 ): string => {
+  const defaultDB = useAsync(() => datasource.getDefaultOrFirstDatabase(), [datasource]);
```
andresmgot, line 143: nit: if you are using this only within the  `useMemo` you can define it there

src/components/QueryEditor.tsx
```diff
@@ -156,11 +157,15 @@ const useSelectedDatabase = (
     }
 
     if (options.length > 0) {
-      return options[0].value;
+      const result = options.find((x) => x.value === defaultDB.value);
+
+      if (result) {
+        return result.value;
+      }
```
andresmgot, line 164: I guess you are missing an `else` here to return the first one if the default has not been found.

andresmgot (commented):
Thanks! Can you add a quick test for this?

src/components/QueryEditor.tsx
asimpson: If I do a `useAsync` inside a `useMemo` that results in a warning about using hooks inside `useMemo`.

andresmgot (approved):
Great, thanks!

andresmgot:
Merging this to prepare a release :+1:

PR #432, opened 2022-07-19
Correctly cast dynamically typed columns

### LINKED ISSUE #429: Dynamic columnn not being casted
**What happened**:

When using a dynamic column in the Group by or Aggregate field, it's necessary to cast the value to a simple type: e.g. `sum(col)` -> `sum(tolong(col))`.

The problem is that the logic, at the moment, to differ is a column is dynamic, is checking if it's part of an object:

https://github.com/grafana/azure-data-explorer-datasource/blob/main/src/KustoExpressionParser.ts#L402
```ts
const isDynamic = (column: string): boolean => {
  return !!(column && column.indexOf('[') > -1) || !!(column && column.indexOf('todynamic') > -1);
};
```

So if a column is a simple value, it won't be detected as dynamic, skiping the casting.

We need to fix that logic.

**Screenshots**

**Environment**:

- Grafana version:
- Plugin version: 4.1.2
- OS Grafana is installed on:
- User OS & Browser:
- Others:

aangelisc:
Fixes #429 where 'simple' dynamic columns were not being appropriately cast in `group by` and `summarize` statements.

### DIFF
--- a/src/KustoExpressionParser.ts
+++ b/src/KustoExpressionParser.ts
@@ -399,16 +399,22 @@ const defaultTimeColumn = (columns?: AdxColumnSchema[], expression?: QueryExpres
   return toType(column.CslType, column.Name);
 };
 
-const isDynamic = (column: string): boolean => {
-  return !!(column && column.indexOf('[') > -1) || !!(column && column.indexOf('todynamic') > -1);
+const isDynamic = (column: string | AdxColumnSchema | undefined): boolean => {
+  if (column && typeof column === 'object') {
+    return column.Type === 'dynamic';
+  } else if (typeof column === 'string') {
+    return !!(column && column.indexOf('[') > -1) || !!(column && column.indexOf('todynamic') > -1);
+  } else {
+    return false;
+  }
 };
 
 const castIfDynamic = (column: string, tableSchema?: AdxColumnSchema[], schemaName?: string): string => {
-  if (!isDynamic(schemaName || column) || !Array.isArray(tableSchema)) {
+  const columnSchema = tableSchema?.find((c) => c.Name === (schemaName || column));
+
+  if (!isDynamic(columnSchema || schemaName || column) || !Array.isArray(tableSchema)) {
     return column;
   }
-
-  const columnSchema = tableSchema.find((c) => c.Name === (schemaName || column));
 
   if (!columnSchema) {
     return column;
--- a/src/KustoExpressionParser.ts
+++ b/src/KustoExpressionParser.ts
@@ -40,7 +40,7 @@ export class KustoExpressionParser {
 
     const parts: string[] = [];
     this.appendProperty(context, query.expression.from, parts);
-    this.appendTimeFilter(context, undefined, parts);
+    this.appendTimeFilter(context, undefined, parts, tableSchema);
 
     const where = replaceByIndex(query.index, query.expression.where, query.search);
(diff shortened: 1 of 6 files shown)


src/KustoExpressionParser.ts
```diff
@@ -399,17 +399,23 @@ const defaultTimeColumn = (columns?: AdxColumnSchema[], expression?: QueryExpres
   return toType(column.CslType, column.Name);
 };
 
-const isDynamic = (column: string): boolean => {
-  return !!(column && column.indexOf('[') > -1) || !!(column && column.indexOf('todynamic') > -1);
+const isDynamic = (column: string | AdxColumnSchema | undefined): boolean => {
+  if (column && typeof column === 'object') {
+    return column.Type === 'dynamic';
+  } else if (typeof column === 'string') {
+    return !!(column && column.indexOf('[') > -1) || !!(column && column.indexOf('todynamic') > -1);
```
andresmgot, line 406: we probably don't need this case anymore, right? And we can simplify the column type to be only `AdxColumnSchema | undefined`

src/datasource.ts
```diff
@@ -383,13 +383,13 @@ const recordSchemaArray = (name: string, types: AdxSchemaDefinition[], result: A
   if (types.every((t) => typeof t === 'string' && toPropertyType(t) === QueryEditorPropertyType.Number)) {
     // If all the types are numbers, the double takes precedence since it has more precission.
     const cslType = types.find((t) => typeof t === 'string' && (t === 'double' || t === 'real')) || defaultCslType;
-    result.push({ Name: name, CslType: cslType as string });
+    result.push({ Name: name, CslType: cslType as string, Type: 'dynamic' });
```
andresmgot, line 386: I don't think we should use the `Type` field since that field is already a used field in a schema and it should be the [".NET equivalent type"](https://docs.microsoft.com/en-us/azure/data-explorer/kusto/query/scalar-data-types/). I would instead extend the AdxColumnSchema type so we can inject some metadata into it:

```tsx
  Name: string;
  CslType: string;
  Type?: string;
  CslDefaultValue?: string;
  // tracking internally if a type is inferred from a dynamic type
  isDynamic?: boolean;
}
```

andresmgot (commented):
Nice solution! I just have a couple of minor suggestions

src/datasource.ts
aangelisc: Good shout 😊

src/KustoExpressionParser.ts
```diff
@@ -399,17 +406,21 @@ const defaultTimeColumn = (columns?: AdxColumnSchema[], expression?: QueryExpres
...
 
-const isDynamic = (column: string): boolean => {
-  return !!(column && column.indexOf('[') > -1) || !!(column && column.indexOf('todynamic') > -1);
+const isDynamic = (column: AdxColumnSchema | undefined): boolean => {
+  if (column && column.isDynamic) {
+    return column.isDynamic;
+  } else {
+    return false;
+  }
 };
```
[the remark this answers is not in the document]
aangelisc: Good catch 😊

andresmgot (approved):
LGTM!

src/datasource.ts
andresmgot, line 386: it seems that you forgot to update these to  use `isDynamic` instead

PR #439, opened 2022-08-03
Query builder: Add support for colums with spaces

### LINKED ISSUE #318: Query builder: Unable to list filter values when the key include spaces
**What happened**:

Got a 400 error when trying to select information from a database table in ADX. My running theory based on the error, is that the ADX connector queries cant load data from an ADX table column with a "space" in the name. See the GIF below for details.
The connector properly loads all other columns.

I expect to see the dropdown list of Truck IDs.

1. Using the ADX connector as a data source (Make sure to have a column with a space)
2. Open the dashboard builder
3. Select the ADX connection as a data source
4. Select the database
5. Select the table
6. Filter "where XXXXX ==" being the column of interest

**Screenshots**
GIF of Error:
Animation

Image of it working normally for a different tag:
image

**Environment**:

- Grafana version: v8.3.5
- Plugin version: 3.6.1
- OS Grafana is installed on: Windows (but using a cloud instance)
- User OS & Browser: Chrome
- Others:

andresmgot:
Started as a mob session, I finished work left.

Screenshot from 2022-08-03 13-15-17

This ended up no being so trivial because there are several corner cases, like queries used to get suggestions (`toAutoCompleteQuery`), columns that apart from having a space are `dynamic` or the additional handling made for macros in the backend.

I think I have added a test for all those cases but let me know if you spot something else!

Fixes #318

### DIFF
--- a/pkg/azuredx/models/macro.go
+++ b/pkg/azuredx/models/macro.go
@@ -35,7 +35,7 @@ func NewMacroData(tr *backend.TimeRange, intervalMS int64) MacroData {
 // macroRE is a regular expression to match available macros
 var macroRE = regexp.MustCompile(`\$__` + // Prefix: $__
 	`(timeFilter|timeFrom|timeTo|timeInterval)` + // one of macro root names
-	`(\([a-zA-Z0-9_\s.-]*?\))?`) // optional () or optional (someArg)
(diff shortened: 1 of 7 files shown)


aangelisc (approved):
LGTM and nice work with the tests 😊

yaelleC (approved):
LGTM. Thanks for finishing up what the mob started 3 weeks ago 😜 !

Issue #436, opened 2022-07-22
ADX data source plugin consuming a large amount of memory


yaelleC:
**What happened**:

Related to [escalation#3328](https://github.com/grafana/support-escalations/issues/3328) 
ADX is consuming too much memory. In case of the escalation, the plugin got OOM killed

**What you expected to happen**:

> ADX plugin should not consistently occupy more than about 500MB memory after serving 2 to 3 ds queries which response size is less than 50MB.

**How to reproduce it (as minimally and precisely as possible)**:

Create an ADX query and try running it for 2-7 days
The `Time Series` format seems to fail more often than the `Table` one 🤷‍♀️ 

**Screenshots**
![Screenshot 2022-07-22 at 12 20 20](https://user-images.githubusercontent.com/42030685/180419256-86cfa526-0bd6-4939-a313-a1df1ef19182.png)

**Anything else we need to know?**:

**Environment**:

- Grafana version: 9.0.5-pre
- Plugin version: 4.0.0 - also tried upgrading 4.1.3 (same issue)
- OS Grafana is installed on: HG
- User OS & Browser: Chrome

andresmgot:
So, a little bit of analysis of what's happening here. The issue is due to the fact that by default, selecting "Format as: Time series" will convert the Frames (we can refer them as rows of a table) from "Long" to "Wide":

https://github.com/grafana/azure-data-explorer-datasource/blob/e25f75a831651e5825bef183496d778c87624f50/pkg/azuredx/datasource.go#L185-L186

This cause that any field of the table that is either a string or a boolean will be treated as a "label", this is a new dimension for the frame. For example, the following rows:

| time | value | field1 | field2 |
| --- | --- | --- | --- |
| 1:00 | 1 | foo | bar |
| 2:00 | 2 | other | value |

`field1` and `field2` will be new dimensions for the time series, which would generate the following response:

| time | value{field1: foo, field2:bar} | value{field1: foo, field2:value} | value{field1: other, field2:bar} | value{field1: other, field2:value} 
| --- | --- | --- | --- | --- |
| 1:00 | 1 | | | |
| 2:00 |  | | | 2 |

Each one of those cells is a point to draw in the graph. As the number of fields and values grow, the number of points to draw grows exponentially.

The current way of avoiding this issue is simply defining a more complete query. If we add a filter `| where field1 == 'foo'` then we limit that growth. The problem is that queries by default will still hit this issue.

Given that we have two options:

[...]

PR #440, opened 2022-08-04
Change default format to Table

andresmgot:
Also fixed a bug to allow to run the default query.

Peek 2022-08-04 16-18

Temporary fix for #436


aangelisc (approved):
LGTM 😊

PR #443, opened 2022-08-09
New query editor skeleton

andresmgot:
This is the first PR for #391 

It includes several things:
 - All of the existing code related to the query editor has been moved under the folder `LegacyQueryEditor`
 - The new editor is behind a feature flag `adxNewEditor`
 - I have bumped the version of the @grafana/ui package to 9.1.0. This is because the latest version of the package includes the content of the previously @grafana/experimental-ui. It may require that we bump the version of the rest of packages. I also need to double check what's the minimal version supported.
 - I have also removed some unused files I found along the way.
 - For the new editor, I have only added the header, which includes the selector of the database and the visual / KQL switch

Peek 2022-08-09 16-26


asimpson (approved):
LGTM, let's go!

PR #444, opened 2022-08-10
New editor: Add format selector

andresmgot:
Ref: #391 

Adding the "Format as" selector, maintaining existing logic. I added it to the header since it's common for both visual and KLQ editor: 

Screenshot from 2022-08-10 16-21-35


aangelisc (approved):
Nice😊

PR #451, opened 2022-08-22
Visual Editor: Table selector

andresmgot:
First component of the visual editor, the table selector:

Screenshot from 2022-08-22 16-53-49

The code is a small adaptation of the one already existing in the LegacyEditor, the only changes are visual.

Additional notes:

I considered changing how the query is structured and avoid the `expression` definition which is a bit complex (basically an `expression` object can define anything, from a table selection to a summarize function). At the end I decided not to go that path for two major reasons: It would be reinventing the wheel, having to re-define the Kusto parser and it would also require to migrate queries based on `expressions` to the new format, which doesn't help with the complexity of the code to maintain.

Also:

The `VisualQueryEditor` component is likely to get huge so I may move this code to another subcomponent in the future. I haven't done that yet since I prefer to have the full layout first.


asimpson (changes requested):
Dig it :trophy:

asimpson (approved):

PR #455, opened 2022-08-25
New editor: Filters

andresmgot:
This PR adds the filter section to the new editor. The biggest change from the previous experience is that now, rather than selecting "AND" or "OR" conditions when adding new filters, users select a new "Filter group" (AND) or a simply a new filter (OR). This is explained in the tooltips.

This is a demo (sorry, I don't know why the cursor is not displayed :thinking:):

https://user-images.githubusercontent.com/4025665/186642991-482e3174-894c-4972-a032-5237655832af.mp4

Please give it a try and let me know if you find any issue! Or if you have any UX suggestion.

It's difficult to review such a big PR so I'll add inline comments to try to make it easier and provide focus points.

### DIFF
--- a/src/KustoExpressionParser.ts
+++ b/src/KustoExpressionParser.ts
@@ -46,9 +46,11 @@ export class KustoExpressionParser {
     this.appendProperty(context, query.expression.from, parts);
     this.appendTimeFilter(context, undefined, parts, tableSchema);
 
-    const where = replaceByIndex(query.index, query.expression.where, query.search);
+    if (query.index) {
+      const where = replaceByIndex(query.index, query.expression.where, query.search);
+      this.appendWhere(context, where, parts, 'where');
+    }
     const column = query.search.property.name;
-    this.appendWhere(context, where, parts, 'where');
 
     parts.push('take 50000');
     parts.push(`distinct ${context.castIfDynamic(column)}`);
--- a/src/components/QueryEditor/VisualQueryEditor.tsx
+++ b/src/components/QueryEditor/VisualQueryEditor.tsx
@@ -8,6 +8,7 @@ import { useAsync } from 'react-use';
 import { AdxSchemaResolver } from 'schema/AdxSchemaResolver';
 import { QueryEditorPropertyDefinition, QueryEditorPropertyType } from 'schema/types';
 import { AdxDataSourceOptions, AdxSchema, defaultQuery, KustoQuery } from 'types';
+import FilterSection from './VisualQueryEditor/FilterSection';
 
 type Props = QueryEditorProps<AdxDataSource, KustoQuery, AdxDataSourceOptions>;
 
@@ -92,6 +93,9 @@ export const VisualQueryEditor: React.FC<VisualQueryEditorProps> = (props) => {
           </EditorField>
         </EditorFieldGroup>
       </EditorRow>
+      <FilterSection {...props} tableSchema={tableSchema} />
+      {/* TODO: Use proper preview component */}
+      <pre>{query.query}</pre>
     </EditorRows>
   );
 };
--- a/src/components/QueryEditor/VisualQueryEditor/FilterItem.test.tsx
+++ b/src/components/QueryEditor/VisualQueryEditor/FilterItem.test.tsx
@@ -1,0 +1,94 @@
+import { render, screen } from '@testing-library/react';
+import { mockDatasource, mockQuery } from 'components/__fixtures__/Datasource';
+import { AdxColumnSchema } from 'types';
+import FilterItem from './FilterItem';
+import React from 'react';
+import { openMenu } from 'react-select-event';
+import { QueryEditorPropertyType } from 'schema/types';
+import userEvent from '@testing-library/user-event';
+
+const defaultProps = {
+  datasource: mockDatasource(),
+  query: mockQuery,
+  filter: {},
+  columns: [],
+  templateVariableOptions: {},
+  onChange: jest.fn(),
+  onDelete: jest.fn(),
+};
+
+describe('FilterItem', () => {
+  it('should select a column', () => {
+    const onChange = jest.fn();
+    const columns: AdxColumnSchema[] = [
+      {
+        Name: 'foo',
+        CslType: 'string',
+      },
+    ];
+    render(<FilterItem {...defaultProps} columns={columns} onChange={onChange} />);
+    const sel = screen.getByLabelText('column');
+    openMenu(sel);
+    screen.getByText('foo').click();
+    expect(onChange).toHaveBeenCalledWith(
+      expect.objectContaining({ property: { name: 'foo', type: QueryEditorPropertyType.String } })
+    );
+  });
+
+  it('should select an operator', () => {
+    const onChange = jest.fn();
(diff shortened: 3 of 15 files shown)


src/KustoExpressionParser.ts
```diff
@@ -46,9 +46,11 @@ export class KustoExpressionParser {
     this.appendProperty(context, query.expression.from, parts);
     this.appendTimeFilter(context, undefined, parts, tableSchema);
 
-    const where = replaceByIndex(query.index, query.expression.where, query.search);
+    if (query.index) {
+      const where = replaceByIndex(query.index, query.expression.where, query.search);
```
andresmgot, line 50: This `replaceByIndex` is meant to get all the `where` clauses from a query, and remove the one that is being edited at the moment, to get a valid query to get suggestions.

The issue is that the new `EditorList` component does not provide the index of each element so I cannot make this replacement. What this means is that, for suggestions, we are not taking current filters to filter out suggestions.

src/components/QueryEditor/VisualQueryEditor.tsx
```diff
@@ -92,6 +93,9 @@ export const VisualQueryEditor: React.FC<VisualQueryEditorProps> = (props) => {
           </EditorField>
         </EditorFieldGroup>
       </EditorRow>
+      <FilterSection {...props} tableSchema={tableSchema} />
+      {/* TODO: Use proper preview component */}
+      <pre>{query.query}</pre>
```
andresmgot, line 98: I plan to add a proper component for this in the future, but in the meantime, this is useful for devs purposes.

src/components/QueryEditor/VisualQueryEditor/FilterItem.tsx
```diff
@@ -0,0 +1,114 @@
...
+        columns
+      )
+      .then((result: string[]) => {
+        if (Array.isArray(templateVariableOptions.options) && filter.property?.type === 'string') {
+          // When selecting a string, automatically quote template variables to generate a valid syntax
+          templateVariableOptions.options = templateVariableOptions.options.map((op: SelectableValue<string>) => {
+            return { label: op.value, value: `'${op.value}'` };
+          });
+        }
+        return result.map((r): SelectableValue<string> => ({ label: r, value: r })).concat(templateVariableOptions);
```
andresmgot, line 61: this logic return suggestions for a filter. In the `search` expression yhou can see that we are trying to find properties that are not empty for a certain value (`property.name`). It append template variables to the final result.

src/components/QueryEditor/VisualQueryEditor/FilterItem.tsx
```diff
@@ -0,0 +1,114 @@
...
+  };
+
+  const [state, loadOptions] = useAsyncFn(loadValues, [query, filter.property]);
+
+  return (
+    <InputGroup>
+      <Select
+        aria-label="column"
+        width="auto"
+        value={filter.property?.name ? valueToDefinition(filter.property?.name) : null}
```
andresmgot, line 72: all these `*ToDefinition` and `*toOption` are just formatting helpers, you can see more details in the definition but it's basically handling labels and types.

src/components/QueryEditor/VisualQueryEditor/FilterItem.tsx
```diff
@@ -0,0 +1,114 @@
...
+      />
+      <Select
+        aria-label="operator"
+        width="auto"
+        value={filter.operator?.name && toOption(filter.operator.name)}
+        options={toOperatorOptions(filter.property?.type)}
+        onChange={({ value }) => value && onChange(setOperatorExpressionName(filter, value))}
+      />
+      <div hidden={filter.property?.type !== QueryEditorPropertyType.Number}>
+        <AutoSizeInput
```
andresmgot, line 85: This logic shows a numeric input if the property is a number. The normal thing here would be to do:

```ts
{filter.property?.type === QueryEditorPropertyType.Number} && <AutoSizeInput
```

But I found out that the component `EditorList` freaks out if children appear / disappear, causing them to be wrongly resized. The only solution I found that worked was to wrap these component in `div`s and hide them.

src/components/QueryEditor/VisualQueryEditor/FilterSection.tsx
```diff
@@ -0,0 +1,77 @@
...
+          </EditorFieldGroup>
+        </EditorRow>
+      ))}
+      <EditorRow>
+        <EditorFieldGroup>
+          <EditorField label="Add filter group" optional={true} tooltip="All filter groups should return true">
+            <Button
+              variant="secondary"
+              icon="plus"
+              onClick={() => {
```
andresmgot, line 53: explanation: clicking in the "+" button, adds a new `expression`to the `where.expressions`. Note that each of those expressions are joined with the "and" operator (in kusto this means different `| where` clauses).

src/components/QueryEditor/VisualQueryEditor/KQLFilter.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+    const where = { ...query.expression.where };
+    if (newItems.length) {
+      (where.expressions[index] as QueryEditorArrayExpression).expressions = validExpressions;
+    } else {
+      // The expression is empty, remove it
+      const expr = where.expressions;
+      expr.splice(index, 1);
+      where.expressions = expr;
+    }
```
andresmgot, line 68: explanation: this code only add new expressions to the query if the expression is complete (e.g. it has the three elements, column, operator and value).

In the case that the `change` event is triggered with removing a condition, and there are no more expressions, we remove the entry.

Note: I have found a bug here. If there are multiple groups, removing the first one messes the output :thinking: 

Peek 2022-08-25 12-54

src/components/QueryEditor/VisualQueryEditor/utils/operators.ts
```diff
@@ -0,0 +1,340 @@
...
+    title: op.Example,
+  }));
+};
+
+export const isMulti = (operator?: string) => {
+  // Some operators can be used to compare multiple values
+  return !!operator && ['in', '!in', 'in~', '!in~', 'has_all', 'has_any'].includes(operator);
+};
+
+export const OPERATORS: { [type: string]: OperatorInfo[] } = {
```
andresmgot, line 27: source of this is https://docs.microsoft.com/en-us/azure/data-explorer/kusto/query/datatypes-string-operators

aangelisc (approved):
Great work on this, especially the testing! Just one nit but good to go 👍

src/components/QueryEditor/VisualQueryEditor/FilterItem.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+      />
+    );
+    const sel = screen.getByLabelText('column value');
+    openMenu(sel);
+    (await screen.findByText('Template Variables')).click();
+    (await screen.findByText('$foo')).click();
+    expect(onChange).toHaveBeenCalledWith(expect.objectContaining({ operator: { name: '==', value: "'$foo'" } }));
+  });
+
+  it('should select a template variable', async () => {
```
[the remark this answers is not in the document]
andresmgot: oops, copy-paste issue :)

PR #460, opened 2022-09-09
New builder: Aggregate

andresmgot:
New section "Aggregate" which generates a summarize expression:

Screenshot from 2022-09-09 16-16-22

Notes:
 - Aggregate functions usually requires a column to apply it to (unless count)
 - Percentile is the only function that requires an additional parameter (percentile number)

### DIFF
--- a/src/components/QueryEditor/VisualQueryEditor/AggregateItem.tsx
+++ b/src/components/QueryEditor/VisualQueryEditor/AggregateItem.tsx
@@ -1,0 +1,131 @@
+import React from 'react';
+
+import { SelectableValue } from '@grafana/data';
+import { Label, Select } from '@grafana/ui';
+import { AccessoryButton, InputGroup } from '@grafana/experimental';
+
+import { AdxDataSource } from '../../../datasource';
+import { AdxColumnSchema, KustoQuery } from '../../../types';
+import {
+  QueryEditorExpressionType,
+  QueryEditorReduceExpression,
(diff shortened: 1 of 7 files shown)


src/components/QueryEditor/VisualQueryEditor/AggregateItem.tsx
```diff
@@ -0,0 +1,131 @@
...
+                  },
+                ],
+                type: QueryEditorExpressionType.Reduce,
+              });
+          }}
+        />
+      )}
+      {aggregate.reduce?.name !== AggregateFunctions.Count && (
+        <>
+          <Label style={{ margin: '9px 9px 0 9px' }}>of</Label>
```
asimpson, line 99: Do we have this pattern else-where? The `of` feels disconnected to me. Might be worth running by UX?

asimpson, line 99: Do we have this pattern else-where? The `of` feels disconnected to me. Might be worth running by UX?

asimpson (approved):
Nice work on this! :trophy:

src/components/QueryEditor/VisualQueryEditor/AggregateItem.tsx
andresmgot: Good shout, for the moment I am only mimicking whatever was already in the previous editor but I am taking note about this as a specific question for UX

PR #462, opened 2022-09-12
Fix default loading

andresmgot:
Noticed that when creating a new query (with the new editor), this error appears:

Screenshot from 2022-09-12 11-54-15

It was because the default database was not being stored in the query object.


yaelleC (approved):
LGTM 🚀

asimpson (approved):
Nice work!

PR #463, opened 2022-09-12
Add run button

andresmgot:
Adding a "Run" button to run queries both in the query builder and the raw editor (removed the onBlur behavior there):

Screenshot from 2022-09-12 12-35-33

Unrealted but added to this PR: avoid showing the confirmation modal changing from raw to builder if the query has not changed.

### DIFF
--- a/src/components/QueryEditor/QueryHeader.test.tsx
+++ b/src/components/QueryEditor/QueryHeader.test.tsx
@@ -102,7 +102,7 @@ describe('QueryEditor', () => {
       const onRunQuery = jest.fn();
       render(<QueryHeader {...defaultProps} schema={schema} onRunQuery={onRunQuery} />);
       await waitFor(() => screen.getByText('foo'));
-      screen.getByRole('button').click();
+      screen.getByText('Run query').click();
       expect(onRunQuery).toHaveBeenCalledTimes(1);
     });
   });
--- a/src/components/QueryEditor/QueryHeader.test.tsx
+++ b/src/components/QueryEditor/QueryHeader.test.tsx
@@ -34,6 +34,7 @@ const defaultProps = {
   schema: defaultSchema,
   dirty: false,
   setDirty: jest.fn(),
+  onRunQuery: jest.fn(),
 };
 
 describe('QueryEditor', () => {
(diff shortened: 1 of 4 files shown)


src/components/QueryEditor/QueryHeader.test.tsx
```diff
@@ -96,6 +97,14 @@ describe('QueryEditor', () => {
       expect(onChange).toHaveBeenCalledWith(expect.objectContaining({ rawMode: false }));
       expect(setDirty).toHaveBeenCalledWith(false);
     });
+
+    it('runs a query', async () => {
+      const onRunQuery = jest.fn();
+      render(<QueryHeader {...defaultProps} schema={schema} onRunQuery={onRunQuery} />);
+      await waitFor(() => screen.getByText('foo'));
+      screen.getByRole('button').click();
```
yaelleC, line 105: I know there is just one button - for now - but maybe we can add a name property to look for the `run` button?

src/components/QueryEditor/QueryHeader.tsx
```diff
@@ -100,6 +101,9 @@ export const QueryHeader = (props: QueryEditorHeaderProps) => {
         }}
       />
       <FlexItem grow={1} />
+      <Button variant="primary" icon="play" size="sm" onClick={() => onRunQuery()}>
```
yaelleC, line 104: 🤔  Any reason why you can't write `onClick={onRunQuery}` ?

yaelleC (approved):
LGTM

Added a couple of minor comments.

Also, it would be nice to get some consistency in the run button. I'll add that to the list of things to discuss next week with UX.
It looks like every data source owner arbitrarily decide on text + style for it 🤷‍♀️ , a few examples:

Screenshot 2022-09-12 at 13 27 55
Screenshot 2022-09-12 at 13 28 14
Screenshot 2022-09-12 at 13 29 15

asimpson (approved):

andresmgot:
> Also, it would be nice to get some consistency in the run button

I agree, it'd be nice to have a component that we can just use so everyone uses the same. My logic here was:

 - The button is quite important since it's the main way of executing the query so I chose the "primary" button approach (with the icon)
 - Saying "run queries" or "run query" is IMO redundant, it's obvious what you are running :sweat_smile: 
 
 I feel less strongly about the second so this is at least consistent with BigQuery

PR #467, opened 2022-09-14
New editor: Style for query preview

andresmgot:
Final component of #391 

Move the preview to its own section and add the capability to show/hide it:

Peek 2022-09-14 12-01

I tried to use a component with some syntax highlighting but:
 - Using a `CodeEditor` (monaco) is a bit heavy. Also it's not possible to set the height automatically which is not very friendly for the preview. Also, the same editor is available when switching to the KQL editor.
 ~~- I also tried to use Prismjs, which is used in other editors but the language for `kusto` is not included by default.~~

### DIFF
--- a/src/components/QueryEditor/VisualQueryEditor.tsx
+++ b/src/components/QueryEditor/VisualQueryEditor.tsx
@@ -12,6 +12,7 @@ import { AdxDataSourceOptions, AdxSchema, defaultQuery, KustoQuery } from 'types
 import FilterSection from './VisualQueryEditor/FilterSection';
 import AggregateSection from './VisualQueryEditor/AggregateSection';
 import GroupBySection from './VisualQueryEditor/GroupBySection';
+import KQLPreview from './VisualQueryEditor/KQLPreview';
 
 type Props = QueryEditorProps<AdxDataSource, KustoQuery, AdxDataSourceOptions>;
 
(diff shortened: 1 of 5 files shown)


src/components/QueryEditor/VisualQueryEditor.tsx
```diff
@@ -99,8 +100,7 @@ export const VisualQueryEditor: React.FC<VisualQueryEditorProps> = (props) => {
       <FilterSection {...props} tableSchema={tableSchema} />
       <AggregateSection {...props} tableSchema={tableSchema} />
       <GroupBySection {...props} tableSchema={tableSchema} />
-      {/* TODO: Use proper preview component */}
```
asimpson, line 102: :raised_hands:

asimpson (approved):
LGTM.

It does look like [Prism added Kusto support](https://github.com/PrismJS/prism/pull/3068), did you try the keyword `kusto`?

asimpson:
Yes! Prism looks great!!! :trophy:

andresmgot:
> It does look like [Prism added Kusto support](https://github.com/PrismJS/prism/pull/3068), did you try the keyword `kusto`?

I saw that but then when I checked how other plugins used Prism, I saw that `Prism.languages` did not include `kusto` so I could not use it that way (I also read somewhere that it was needed to modify webpack to include it). Anyway, now I've found this snippet: https://gist.github.com/awran5/b2fb99b91d279b700aff6d3a0f1f223d and it seems to work! Default themes are not mindblowing but we have now some syntax highlighting: 

Screenshot from 2022-09-14 16-50-41

asimpson:
>Anyway, now I've found this snippet: https://gist.github.com/awran5/b2fb99b91d279b700aff6d3a0f1f223d and it seems to work! Default themes are not mindblowing but we have now some syntax highlighting:
> 
 Ah, slick. Nice find! :art:

PR #468, opened 2022-09-15
New editor: fix clean up when changing table

andresmgot:
Fix the editor to clean up the rest of fields when the table or the database changes

Peek 2022-09-15 17-00

Peek 2022-09-15 17-01


asimpson (approved):
Looks good, tests pass, and works as expected :trophy:

Issue #454, opened 2022-08-25
Choose correct dynamic type when filters are applied


aangelisc:
See grafana/support-escalations#3478. When a filter is applied on a column with a dynamic type the schema used is for the full column. This can lead to the incorrect type being chosen when the data is cast in a summarize statement.

Expected behaviour:

The appropriate type should be chosen based on the remaining values after a filter is applied to the column.

yaelleC:
We might need UX input there to decide on solution for the issue

andresmgot:
Let me elaborate in the issue so we can decide on specific solutions.

The root of the issue is that a dynamic column needs to be parsed as a specific type so you cannot do:

```
| where DynamicCol > 1
```

Instead it should be casted:

```
| where tolong(DynamicCol) > 1
```

To know which specific types a dynamic column can have, we are getting the schema using the `buildschema` function from ADX. The schema is obtained once, and it ignores the current query. The final query executed in `getDynamicSchema` is:

```
MyTable
 | take 50000
 | where isnotnull(DynamicCol)
 | project DynamicCol
 | summarize buildschema(DynamicCol)
```

This returns the list of all the different types from the last 50000 rows of the table. The list can contain _any_ type (e.g. `[long, double, string]`. When all the types are number types, we automatically choose `double` as the type of column since it's the one with more precission. If the type can be either a number or another thing (string, boolean...) we cannot make that assumption since converting a string to a double would return an incorrect value. In that case, we just select the first type reported from the list.

Give that, there are several things we can do to mitigate the issue:

1. (effort/small) Change the condition in which we select the first type and always use `double` if it's an option. This would unblock https://github.com/grafana/support-escalations/issues/3478 and it's not necessarily more wrong than the current choice. Also note that doing `todouble("foo")` does not error, it simply return an empty value.
2. (effort/medium) Update the stored schema based on the query filters. This would narrow down the scope of the schema resolution query, potentially filtering out spurious values. Note that this is not a final solution since it can still be the case that uncompatible types may be returned. The final query would look something like:

```
MyTable
| where $__timeFilter(TimeCol)
| where Col1 == 'value1'
| where Col2 == 'value2'
| take 50000
| where isnotnull(DynamicCol)
| project DynamicCol
| summarize buildschema(DynamicCol)
```

[...]

PR #471, opened 2022-09-16
Dynamic columns: Cast to double if type exists in schema

yaelleC:
## Description

Step one of: https://github.com/grafana/azure-data-explorer-datasource/issues/454
To fix an escalation we received with the wrong type being used in cast.

This PR introduces a check for `double` in the types list of a dynamic column, if it is present we use it to cast, otherwise we default to the current behaviour: using the first type on the list.

This PR is not the perfect fix, but as explained on the issue, casting to double if it exists makes more sense than arbitrarily taking the first type in array.

Next steps will involve updating the stored schema based on the query filters to refine the type used in cast.

### DIFF
--- a/src/KustoExpressionParser.test.ts
+++ b/src/KustoExpressionParser.test.ts
@@ -511,7 +511,7 @@ describe('KustoExpressionParser', () => {
       );
     });
 
-    it('should parse expression with summarize of sum on dynamic column', () => {
+    it('should parse expression with summarize of sum with cast to double on dynamic column with integer and double types', () => {
       const expression = createQueryExpression({
         from: createProperty('StormEvents'),
         where: createArray([createOperator('column["isActive"]', '==', true)]),
@@ -524,10 +524,17 @@ describe('KustoExpressionParser', () => {
           CslType: 'int',
           isDynamic: true,
         },
+        {
+          Name: 'column["level"]["active"]',
+          CslType: 'double',
+          isDynamic: true,
+        },
       ];
 
       expect(parser.toQuery(expression, tableSchema)).toEqual(
-        'StormEvents' + '\n| where column["isActive"] == true' + `\n| summarize sum(toint(column["level"]["active"]))`
+        'StormEvents' +
+          '\n| where column["isActive"] == true' +
+          `\n| summarize sum(todouble(column["level"]["active"]))`
       );
     });
 
--- a/src/KustoExpressionParser.test.ts
+++ b/src/KustoExpressionParser.test.ts
@@ -511,7 +511,7 @@ describe('KustoExpressionParser', () => {
       );
     });
 
-    it('should parse expression with summarize of sum with cast to double on dynamic column with integer and double types', () => {
+    it('should parse expression with summarize of sum on dynamic column', () => {
       const expression = createQueryExpression({
         from: createProperty('StormEvents'),
         where: createArray([createOperator('column["isActive"]', '==', true)]),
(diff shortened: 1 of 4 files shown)


src/KustoExpressionParser.test.ts
```diff
@@ -524,10 +524,17 @@ describe('KustoExpressionParser', () => {
           CslType: 'int',
           isDynamic: true,
         },
+        {
+          Name: 'column["level"]["active"]',
+          CslType: 'double',
+          isDynamic: true,
+        },
```
andresmgot, line 531: I believe this is not represent reality. The schema will contain only one element per `Name` so, there won't be a repeated `column["level"]["active"]`.

The fix  should be in datasource.ts:

```ts
const recordSchemaArray = (name: string, types: AdxSchemaDefinition[], result: AdxColumnSchema[]) => {
  // If a field can have different types (e.g. long and double)
  // we select the first, assuming they are interchangeable
  const defaultCslType = types[0];
  if (types.every((t) => typeof t === 'string' && toPropertyType(t) === QueryEditorPropertyType.Number)) {
    // If all the types are numbers, the double takes precedence since it has more precission.
    const cslType = types.find((t) => typeof t === 'string' && (t === 'double' || t === 'real')) || defaultCslType;
    result.push({ Name: name, CslType: cslType as string, isDynamic: true });
  } else {
    console.warn(`schema ${name} may contain different types, assuming ${defaultCslType}`);
    if (typeof defaultCslType === 'object') {
      recordSchema(name, types[0], result);
    } else {
      result.push({ Name: name, CslType: defaultCslType, isDynamic: true });
    }
  }
};
```

here is where we record the schema in case the type has multiple values. We just need to remove the condition that checks if all the types are `Number` and simply set the type to double or real if it's part of the list.

yaelleC: I see, so when we get to `escapeAndCastIfDynamic` the selection of a single type has already been made, my bad!

I'll make the changes to `datasource.ts` 🙇‍♀️

src/datasource.ts
```diff
@@ -378,12 +378,14 @@ const dynamicSchemaParser = (frames: DataFrame[]): Record<string, AdxColumnSchem
...
-  // we select the first, assuming they are interchangeable
-  const defaultCslType = types[0];
-  if (types.every((t) => typeof t === 'string' && toPropertyType(t) === QueryEditorPropertyType.Number)) {
-    // If all the types are numbers, the double takes precedence since it has more precission.
-    const cslType = types.find((t) => typeof t === 'string' && (t === 'double' || t === 'real')) || defaultCslType;
-    result.push({ Name: name, CslType: cslType as string, isDynamic: true });
+  // we select double if it exists as it's the one with more precision, otherwise we take the first
+  const defaultCslType = types.find((t) => typeof t === 'string' && (t === 'double' || t === 'real')) || types[0];
+  if (
+    types.length < 2 ||
```
yaelleC, line 384: I understand that if `cslType` is an array it will have more than one value, but I figured it would be good to catch the single value array case to avoid triggering a warning unnecessarily.

Plus it might come in handy when we make the changes to update the stored schema based on the query filters.

andresmgot: I believe that `buildschema` does not return an array if there is only one type but it doesn't hurt to have this check here.

andresmgot: also, just a minor styling thing  but now I would order this a bit different. Something like:

```ts
    if (length > 1 && !allNumbers) { console.warn... }
    // previus else, no need to cast defaultCslType as string
    if (typeof defaultCslType === 'object') {
      recordSchema(name, types[0], result);
    } else {
      result.push({ Name: name, CslType: defaultCslType, isDynamic: true });
    }
```

and then you do the rest o

andresmgot (approved):
LGTM!

PR #470, opened 2022-09-16
Update node deps

andresmgot:
Making levitate happy

PR #472, opened 2022-09-19
Update @grafana packages

andresmgot:
In https://github.com/grafana/azure-data-explorer-datasource/pull/470/ I hit a wall trying to update all deps, which requires to upgrade `jest` but `@grafana/toolkit` is not compatible, yet, with `jest@28`. This PR only upgrades the rest of @grafana deps.


yaelleC (approved):
Thanks!

PR #474, opened 2022-09-19
New editor: Add column selector

andresmgot:
The goal of this PR is to give a final solution (or at least start it) for #436

Now, people can select a subset of columns so only those fields will be processed. This is highly useful when using the time series format (as explained in #436). If some colums are selected, there rest are removed from the possible suggestions in the rest of fields:

Peek 2022-09-19 16-21

## UX open question

In #436 we discussed, that since the "Time series" format was selected by default, it would be nice to automatically select two columns so people don't find that issue: one for a time column and one for numeric value. The workaround was to select "Table" by default but I started to implemented the suggested behavior. I stoped going that way because I am not sure it's the best approach:

 - It assumes that the user is planning to use the time series visualization. That could have been mostly the case in the past but now there are many possible visualization so doing that assumption seems incorrect.
 - It would potentially choose the wrong columns. In that case, the user would need to _delete_ things from the default query and add new items (rather than seeing everything at first glance).
 - If the column does not have a timestamp and a numeric value, choosing the "time series" format by default would be wrong.
 - If the user start adding columns, without noticing the "time series" format (or without knowing the effect), it's more likely that they will hit the issue described in #436.

Because of that, I think the current behavior is more friendly (selecting all columns, with the table format) but let me know if you have any thoughts there!

### DIFF
--- a/src/components/QueryEditor/VisualQueryEditor/TableSection.tsx
+++ b/src/components/QueryEditor/VisualQueryEditor/TableSection.tsx
@@ -7,7 +7,7 @@ import { AdxColumnSchema, AdxDataSourceOptions, defaultQuery, KustoQuery } from 
 import { QueryEditorPropertyDefinition, QueryEditorPropertyType } from 'schema/types';
 import { Select } from '@grafana/ui';
 import { AdxDataSource } from 'datasource';
-import { uniq } from 'lodash';
+import { toColumnNames } from './utils/utils';
 
 type Props = QueryEditorProps<AdxDataSource, KustoQuery, AdxDataSourceOptions>;
 
@@ -73,7 +73,7 @@ const TableSection: React.FC<TableSectionProps> = ({
             aria-label="Columns"
             isMulti
             value={query.expression.columns?.columns ? query.expression.columns.columns : []}
-            options={uniq(tableSchema.value?.map((c) => c.Name.split('[')[0])).map((c) => ({ label: c, value: c }))}
+            options={toColumnNames(tableSchema.value || []).map((c) => ({ label: c, value: c }))}
             placeholder="All"
             onChange={(e) => {
               onChange({
--- a/src/components/QueryEditor/VisualQueryEditor/TableSection.tsx
+++ b/src/components/QueryEditor/VisualQueryEditor/TableSection.tsx
@@ -1,0 +1,97 @@
+import { QueryEditorProps, SelectableValue } from '@grafana/data';
+import { EditorField, EditorFieldGroup, EditorRow } from '@grafana/experimental';
+import { QueryEditorExpressionType } from 'components/LegacyQueryEditor/editor/expressions';
+import React, { useEffect } from 'react';
+import { AsyncState } from 'react-use/lib/useAsyncFn';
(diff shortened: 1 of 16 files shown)


yaelleC:
At first glance a few comments, without looking at the code:

* In your screenshot, when you add a new filter group it somehow creates a new `Filter Group` input above the existing `Add filter group` one... I don't suppose that's expected
* If we default to `all` columns, isn't it more likely that will will hit the memory issue ? Any way we could add default columns, I don't suppose they will always have `timegenerated` or `id`?

andresmgot:
> * In your screenshot, when you add a new filter group it somehow creates a new `Filter Group` input above the existing `Add filter group` one... I don't suppose that's expected?

Yes, that's expected (or at least I coded it that way :P). It's something I also want to check with UX. It's difficult to differentiate between filters of the same group and different filter groups. My other option was to substitue the section that only serves as a button as only a button (said that way it makes sense), but I was not convinced about breaking the homogeneity of the editor. Anyway it's something that we can check with UX, and have some options.

> * If we default to `all` columns, isn't it more likely that will will hit the memory issue ? Any way we could add default columns, I don't suppose they will always have `timegenerated` or `id`? But if they did it might be good to default to these, and have the user add colums as needed instead?

In theory, people won't hit the memory issue when using the "Table" format. What we could do (as talked offline), for the "time series" case, is to render other selectors so it's clear that for time series you need a time column, a number column and optionally other fields. In any case, we can do that as a follow up PR (to avoid adding more changes here).

src/components/QueryEditor/VisualQueryEditor/TableSection.tsx
```diff
@@ -0,0 +1,97 @@
...
+              });
+            }}
+          />
+        </EditorField>
+        <EditorField label="Columns" tooltip={'Select a subset of columns for faster queries'}>
+          <Select
+            aria-label="Columns"
+            isMulti
+            value={query.expression.columns?.columns ? query.expression.columns.columns : []}
+            options={uniq(tableSchema.value?.map((c) => c.Name.split('[')[0])).map((c) => ({ label: c, value: c }))}
```
asimpson, line 76: Nit: worth pulling this out into a standalone method for clarity/legibility reasons?

src/components/QueryEditor/VisualQueryEditor/utils/utils.ts
```diff
@@ -195,3 +197,14 @@ export function sanitizeGroupBy(expression: QueryEditorGroupByExpression): Query
...
 }
+
+export function filterColumns(
+  tableSchema?: AdxColumnSchema[],
+  expression?: QueryEditorColumnsExpression
+): AdxColumnSchema[] | undefined {
+  return expression?.columns?.length
+    ? // filter columns with the same name or under the same dynamic column
+      // e.g. MyCol or MyCol["Inner"]
+      tableSchema?.filter((c) => expression?.columns?.includes(c.Name.split('[')[0]))
```
asimpson, line 208: I can't figure out a clearer way to express this without looping twice :sweat_smile:, nice work!

asimpson (approved):
LGTM :trophy: :rocket: 

- tests pass
- works as expected in browser
- excited to see the memory impact :crossed_fingers:

yaelleC (approved):
LGTM, Thanks !

PR #475, opened 2022-09-20
New editor: Improve filter style

andresmgot:
My goal here is to improve how filters are created using the query builder. I guess it's important to know what's in the legacy editor at the moment since this is what users are use to:

https://user-images.githubusercontent.com/4025665/191294984-f0247f82-a0bc-4768-b5e8-47f3027d1505.mp4

As you can see, when adding filters, the builder ask to choose between "OR", which adds a new condition, and "AND", which addds a new `where` clause.

From that, the new UI is currently as follows:

https://user-images.githubusercontent.com/4025665/191295189-945bcb07-2d0a-4752-8f1d-cc88a603a3cb.mp4

But that has some problems:

 - The "Add filter group" is represented as a section but it's mostly serving as a button.
 - It may not be clear for uses what's the relation between a filter group and a inline filter (OR / AND logic).

To address that I am suggesting the following:

https://user-images.githubusercontent.com/4025665/191295151-fbafbcb3-ab8b-4cc3-b24d-a773c69d548f.mp4

 - Only one section, with a button to add new "groups"
 - Specify in the UI what's the relation between filters (OR / AND)
 - (con) Now the "Filters" section breaks the homogeneus style but it's probably worth it.

Thoughts? @diegoadams @catherineymgui

### DIFF
--- a/src/components/QueryEditor/VisualQueryEditor/KQLFilter.tsx
+++ b/src/components/QueryEditor/VisualQueryEditor/KQLFilter.tsx
@@ -24,6 +24,10 @@ interface KQLFilterProps {
   onChange: (query: KustoQuery) => void;
 }
 
+export interface FilterExpression extends QueryEditorOperatorExpression {
+  index: number;
+}
+
 const KQLFilter: React.FC<KQLFilterProps> = ({
   index,
   query,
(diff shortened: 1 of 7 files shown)


src/components/QueryEditor/VisualQueryEditor/KQLFilter.tsx
```diff
@@ -24,6 +24,10 @@ interface KQLFilterProps {
   onChange: (query: KustoQuery) => void;
 }
 
+export interface FilterExpression extends QueryEditorOperatorExpression {
+  index: number;
+}
```
andresmgot, line 29: Note for the reviewer: I had to define this new type because `EditorList` from @grafana/ui does not include the index of the element being rendered. My plan is to open an issue / PR there to add that but in the meantime this is necessary to know the possition of the filter (and conditionally render the "OR" label).

src/components/QueryEditor/VisualQueryEditor/KQLFilter.tsx
```diff
@@ -34,21 +38,24 @@ const KQLFilter: React.FC<KQLFilterProps> = ({
 }) => {
   // Each expression is a group of several OR statements
   const expressions = (query.expression.where.expressions[index] as QueryEditorArrayExpression)?.expressions;
```
aangelisc, line 40: Why not carry the map to add each index here?

aangelisc (approved):
This looks great to me, I think it's a much improved UI & UX 😊

aangelisc:
A very very small nit: 
It may be worth adding some more padding to avoid this overlap of the selectors.
image

andresmgot:
> It may be worth adding some more padding to avoid this overlap of the selectors.

I think this is a visual effect, it doesn't really overlap, it's just the "focus" line the one that gets behind the selector you hover other. The alternative would be to add some horizontal margin there but it would break the style of the continuous selectors :thinking:. It may be an interessting thing to try out in a hacking session.

PR #476, opened 2022-09-21
Pre-select columns

andresmgot:
Follow up of https://github.com/grafana/azure-data-explorer-datasource/pull/474 (will be kept as a draft until that PR is merged).

To avoid as much as possible the crash of the browser when using the query builder, this PR pre-selects two columns when writing a new query (it doesn't affect existing queries). This affects only the "Time series" format. There is some more information in the tooltip about time series format.

[Screencast from 26-09-22 17:46:13.webm](https://user-images.githubusercontent.com/4025665/192322104-bfe27115-2a6b-4a2e-9c23-f6a5571cb49c.webm)

### DIFF
--- a/src/components/QueryEditor/VisualQueryEditor/utils/utils.ts
+++ b/src/components/QueryEditor/VisualQueryEditor/utils/utils.ts
@@ -1,14 +1,16 @@
 import { SelectableValue, toOption } from '@grafana/data';
 import {
+  QueryEditorArrayExpression,
   QueryEditorColumnsExpression,
   QueryEditorExpressionType,
   QueryEditorGroupByExpression,
   QueryEditorOperatorExpression,
   QueryEditorReduceExpression,
 } from 'components/LegacyQueryEditor/editor/expressions';
-import { isUndefined, uniq } from 'lodash';
+import { intersection, isUndefined, uniq } from 'lodash';
+import { toPropertyType } from 'schema/mapper';
 import { QueryEditorOperatorValueType, QueryEditorPropertyType } from 'schema/types';
-import { AdxColumnSchema } from 'types';
+import { AdxColumnSchema, QueryExpression } from 'types';
 import { AggregateFunctions } from '../AggregateItem';
 import { FilterExpression } from '../KQLFilter';
 import { isMulti, OPERATORS } from './operators';
@@ -219,4 +221,53 @@ export function filterColumns(
       // e.g. MyCol or MyCol["Inner"]
       tableSchema?.filter((c) => expression?.columns?.includes(toColumnName(c)))
     : tableSchema;
+}
+
+// return columns in use by an expression. If none has been specified, return the first
+// time and number columns from the table definition
+export function defaultTimeSeriesColumns(expression: QueryExpression, tableColumns: AdxColumnSchema[]): string[] {
+  const res: string[] = [];
+  if (expression.where.expressions?.length) {
+    (expression.where.expressions as QueryEditorArrayExpression[]).forEach((exp) => {
+      if (exp.expressions.length) {
+        (exp.expressions as QueryEditorOperatorExpression[]).forEach((e) => {
+          if (!res.includes(e.property.name)) {
+            res.push(e.property.name);
+          }
+        });
+      }
+    });
+  }
+  if (expression.reduce.expressions?.length) {
+    (expression.reduce.expressions as QueryEditorReduceExpression[]).forEach((exp) => {
+      if (!res.includes(exp.property.name)) {
+        res.push(exp.property.name);
+      }
+    });
+  }
+  if (expression.groupBy.expressions?.length) {
+    (expression.groupBy.expressions as QueryEditorGroupByExpression[]).forEach((exp) => {
+      if (!res.includes(exp.property.name)) {
+        res.push(exp.property.name);
+      }
+    });
+  }
+
+  const timeCols = tableColumns
+    .filter((cc) => toPropertyType(cc.CslType) === QueryEditorPropertyType.DateTime)
+    .map((c) => c.Name);
+  if (timeCols.length && intersection(res, timeCols).length === 0) {
+    // No time column in use, add the first one
+    res.push(timeCols[0]);
+  }
+
+  const valCols = tableColumns
+    .filter((cc) => toPropertyType(cc.CslType) === QueryEditorPropertyType.Number)
+    .map((c) => c.Name);
+  if (valCols.length && intersection(res, valCols).length === 0) {
+    // No value column in use, add the first one
+    res.push(valCols[0]);
+  }
+
+  return res;
 }
--- a/src/components/QueryEditor/VisualQueryEditor/utils/utils.ts
+++ b/src/components/QueryEditor/VisualQueryEditor/utils/utils.ts
@@ -253,17 +253,23 @@ export function defaultTimeSeriesColumns(expression: QueryExpression, tableColum
     });
   }
 
-  const timeCols = tableColumns
-    .filter((cc) => toPropertyType(cc.CslType) === QueryEditorPropertyType.DateTime)
-    .map((c) => c.Name);
+  const timeCols = tableColumns.reduce<string[]>((cols, col) => {
+    if (toPropertyType(col.CslType) === QueryEditorPropertyType.DateTime) {
+      cols.push(col.Name);
+    }
+    return cols;
+  }, []);
   if (timeCols.length && intersection(res, timeCols).length === 0) {
     // No time column in use, add the first one
     res.push(timeCols[0]);
   }
 
-  const valCols = tableColumns
-    .filter((cc) => toPropertyType(cc.CslType) === QueryEditorPropertyType.Number)
-    .map((c) => c.Name);
+  const valCols = tableColumns.reduce<string[]>((cols, col) => {
+    if (toPropertyType(col.CslType) === QueryEditorPropertyType.Number) {
+      cols.push(col.Name);
+    }
+    return cols;
+  }, []);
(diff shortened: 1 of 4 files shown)


yaelleC:
Like you I really like the pre-selection, I think that is what ultimately will fix the memory issue 👍 

I understand your concerns on the second item, could you add another time column in the `other dimensions` field (for those edge cases we did not think of)? and what about `Player.score` style columns, could it go there?

Overall, if the `other dimensions` fixes the edge cases I would not mind keeping both your changes, if someone really sees the value in having them as separate inputs.
But if it's not clear whether or not the separate fields bring any value, then keeping both `table` and `time series` view consistent would be my first choice too. 

Another idea: if the issue was an error when there is no time and/or number columns selected maybe we could add that as input validation for the `Columns` multiselect in the case of `time series` format? Instead of a separate UI?

sunker:
I think it looks good! A few questions here. 
* Are you saving two complete queries in the query model - one for table and one for time series?
* Why auto select two fields in the table mode? Seems a bit arbitrary to me. User may wonder why these fields were selected and other weren't. I think I'd prefer if none of them were selected by default.
* Same for time series. I think it would make sense to let the user pick the time, value and dimension columns. 
* For the value field, are you just listing the numeric columns? Even though it may be helpful to the user, it sounds a bit risky to me. For example, the timeseries viz accepts booleans as value fields. I'm not sure what other panels accepts, but feels like they could also accept other field types. Maybe list all columns here? Or at least have it accept custom values. 
* Is dimension a concept in the ADX world? If not, may be better to use the work "label"
* Think it's really important to add tooltips for time, value and dimensions column. And also describe this in the docs. Potentially also link to https://grafana.com/docs/grafana/latest/developers/plugins/data-frames/

sunker:
>It's a bit strict to force one and only one time and number columns. It's likely we are going to miss some use case, for example: does it makes sense to select several time columns to do some calculations afterwards?. There is also the caveat that dynamic objects do not play well with this approach, it would not possible to use for example Players.Score as the number column.

You could be right here. For example, I don't know much about the "ADX time series" format. Maybe you can run this by Gabi?

andresmgot:
Thanks a lot for all this feedback @yaelleC @sunker!

> I understand your concerns on the second item, could you add another time column in the `other dimensions` field (for those edge cases we did not think of)?

I didn't add that to this draft but it's tricky if we were to allow it. If you select "TimeGenerated" as the time column and "StartTime" as an additional label (aka dimension) but then you aggregate by "StarTime", suddenly the plugin would use that one (without updating the time column). That's why I left it out, I assume that you don't need anything related to timestamps as labels but I may be wrong.

> and what about `Player.score` style columns, could it go there?

In this case, the column will be just `Players`. To be fair, the capabilities of the builder for dynamic nested objects is limited already. For example, there is no way to just extract the values of `Player.score` without any processing. That's why I just didn't include those "columns" in the value column.

> Overall, if the `other dimensions` fixes the edge cases I would not mind keeping both your changes, if someone really sees the value in having them as separate inputs. But if it's not clear whether or not the separate fields bring any value, then keeping both `table` and `time series` view consistent would be my first choice too.

Yes, my main concern is that we will be limiting stuff in the "time" and "value" columns.
 
> Another idea: if the issue was an error when there is no time and/or number columns selected maybe we could add that as input validation for the `Columns` multiselect in the case of `time series` format? Instead of a separate UI?

You already get that from the current UI (a warning and the message in the panel):

Screenshot from 2022-09-22 12-02-32

> * Are you saving two complete queries in the query model - one for table and one for time series?

No, the query model is the same, it's just UI changes.

> * Why auto select two fields in the table mode? Seems a bit arbitrary to me. User may wonder why these fields were selected and other weren't. I think I'd prefer if none of them were selected by default.

Fair point. It probably makes sense to only pre-select things for the time-series format.

> * Same for time series. I think it would make sense to let the user pick the time, value and dimension columns.

The thing is that if they don't and they click in the "Run Query" button, the browser may crash (or behave super slow). This is the issue described in https://github.com/grafana/azure-data-explorer-datasource/pull/474 and what we want to avoid.

> * For the value field, are you just listing the numeric columns? Even though it may be helpful to the user, it sounds a bit risky to me. For example, the timeseries viz accepts booleans as value fields. I'm not sure what other panels accepts, but feels like they could also accept other field types. Maybe list all columns here? Or at least have it accept custom values.

Yes, the thing I don't really know what makes sense there. It may makes sense to even use another types like timestamps if they are later processed so that's why I am not so sure about adding that field.

> * Is dimension a concept in the ADX world? If not, may be better to use the work "label"

Not really, I don't know why that sounded more familiar to me but label is probably a better name.

> * Think it's really important to add tooltips for time, value and dimensions column. And also describe this in the docs. Potentially also link to https://grafana.com/docs/grafana/latest/developers/plugins/data-frames/

Good point.

-------------

So based on this feedback, I think a good compromise for this may be:

 - For the table format: Do nothing. Let the user select columns, if any. In other case use all the columns since it won't hit the issue in https://github.com/grafana/azure-data-explorer-datasource/pull/474.
 - For the time series format: Pre-select a time and a value column by default but keep the same selector than for the table format. This will prevent people from easily hitting #474 but keeping the flexibility for them to pick any column they want to for any field.
 - Add more information to the tooltip of this field explaining why the pre-selected columns and a link to the docs.

Does this make sense?

sunker:
> Does this make sense?

Yes makes sense to me.

> The thing is that if they don't and they click in the "Run Query" button, the browser may crash (or behave super slow). This is the issue described in https://github.com/grafana/azure-data-explorer-datasource/pull/474 and what we want to avoid.

Sounds like this can be easily prevented by in the datasource file filtering out queries that don't have fields specified?

andresmgot:
> Sounds like this can be easily prevented by in the datasource file filtering out queries that don't have fields specified?

It's tricky because existing queries don't have that field specified so we would need some mitration code. It would be difficult to differ between new queries that forgot to specify this field and new queries that should specify it. Also, not specifying fields would be fine for tables that don't have many columns, with scarce data or if the target time frame is small enough.

sunker:
Can't you check the plugin version? Version x.x.x where format is timeseries and time field and value field is missing should be left out?

andresmgot:
> Can't you check the plugin version? Version x.x.x where format is timeseries and time field and value field is missing should be left out?

The version is not part of the query, right? In any case, it's not that those queries should be forbidden but they have the risk of being super slow. Maybe we can add a warning, which is softer, and can apply to any version :thinking:

yaelleC:
I agree with your 3 points : no preselection for table, preselect in multiselect input for time series and add info to tooltip and doc 👍

sunker:
> The version is not part of the query, right?

No but the plugin version is available in the datasource file, so it would be a simple check. 

>In any case, it's not that those queries should be forbidden but they have the risk of being super slow. Maybe we can add a warning, which is softer, and can apply to any version

This would work. I guess I'm just not a fan of running queries automatically without user interaction. But we do that in many of our plugins, so it's a general problem.

andresmgot:
this is now ready to be reviewed :+1:

src/components/QueryEditor/VisualQueryEditor/utils/utils.ts
```diff
@@ -220,3 +222,52 @@ export function filterColumns(
       tableSchema?.filter((c) => expression?.columns?.includes(toColumnName(c)))
     : tableSchema;
 }
+
+// return columns in use by an expression. If none has been specified, return the first
+// time and number columns from the table definition
+export function defaultTimeSeriesColumns(expression: QueryExpression, tableColumns: AdxColumnSchema[]): string[] {
+  const res: string[] = [];
+  if (expression.where.expressions?.length) {
+    (expression.where.expressions as QueryEditorArrayExpression[]).forEach((exp) => {
```
sunker, line 231: Is this safe? I assume we'll add betterer to the plugins CI sometime soon, so you might as well use type guards now so you don't have to fix it later.

sunker (approved):
Nice!

src/components/QueryEditor/VisualQueryEditor/utils/utils.ts
```diff
@@ -220,3 +222,52 @@ export function filterColumns(
...
+    (expression.groupBy.expressions as QueryEditorGroupByExpression[]).forEach((exp) => {
+      if (!res.includes(exp.property.name)) {
+        res.push(exp.property.name);
+      }
+    });
+  }
+
+  const timeCols = tableColumns
+    .filter((cc) => toPropertyType(cc.CslType) === QueryEditorPropertyType.DateTime)
+    .map((c) => c.Name);
```
aangelisc, line 258:
```suggestion
  const timeCols = tableColumns
    .reduce<string[]>((cols, option) => {
      if (toPropertyType(option.CslType) === QueryEditorPropertyType.DateTime) {
        cols.push(option.Name);
      }
      return cols;
    }, [])
```
Maybe use `reduce` here and below to avoid looping twice?

aangelisc (approved):
Great work! A couple suggestions in the comments but LGTM 😊

src/components/QueryEditor/VisualQueryEditor/utils/utils.ts
andresmgot: It is, the types of the expressions are a bit of a mess at the moment. It's using wrong definitions that are generally bypassed by legacy code. Let me create a follow up issue to refactor this

src/components/QueryEditor/VisualQueryEditor/utils/utils.ts
andresmgot: usually I don't like `reduce` loops that much because they are more difficult to read (and usually arrays are not that big to make a difference) but since here I am doing the same twice (4 loops in total), it's worth it :+1:

PR #479, opened 2022-09-26
Update flow for variable interpolation

### LINKED ISSUE #370: Template variable values with parenthesis result in 400 bad request
**What happened**:
When the template variable value has parenthesis included (e.g. "Product A (legacy), the value becomes malformed when inserted into the query. This results into a 400 bad request. 

No malformed template variable values so that it does not result into a bad request, but a successful request.

1. I created a template variable and used the following query:
```
datatable(municipality:string)
["'s-Gravenhage", "Product l'ecole", "Product A", "Product A (legacy)", "Product B", "Product B (legacy)"];
```
2. I created a table panel and added this query:
```
datatable(field_1:string)
["'s-Gravenhage", "Product l'ecole", "Product A", "Product A (legacy)", "Product B", "Product B (legacy)"]
| where $__contains(field_1, $variables_all)
| project field_1
```
3. I then select a few values from the template variable/filter:
image
4. I then run the query and the 400 bad request appears:
image
When I click the raw query button, I see the following:
image
image

No.

**Environment**:

- Grafana version: 8.5.0
- Plugin version: 3.7.1
- OS Grafana is installed on: Linux (ubuntu) and MacOS
- User OS & Browser: Brave / Google Chrome / Safari / Firefox
- Others:

aangelisc:
Fixes #370 

- Update flow for variable interpolation to ensure template variables in the `$__contains` macro are parsed correctly
- Update tests to cover potential edge cases
- Add autosuggestion for `$__contains` macro

### DIFF
--- a/src/components/QueryEditor/Suggestions.ts
+++ b/src/components/QueryEditor/Suggestions.ts
@@ -63,6 +63,32 @@ export function getFunctions(variables: VariableModel[]): Record<string, AdxFunc
       InputParameters: [],
       OutputColumns: [],
     },
+    $__contains: {
+      Name: '$__contains',
+      Body: `{ colName in ('value1','value2') }`,
+      FunctionKind: 'Unknown',
+      DocString:
+        '##### Used with multi-value template variables.\n\n' +
+        'If `$myVar` has the value `value1`,`value2`, it expands to: `colName in (value1,value2)`.' +
+        'If using the `All` option, then check the `Include All Option` checkbox and in the `Custom all value` field type in the value `all`. If `$myVar` has value `all` then the macro will instead expand to `1 == 1`.' +
+        'For template variables with a lot of options, this will increase the query performance by not building a large "where..in" clause.' +
+        '[Grafana docs](https://grafana.com/docs/grafana/latest/datasources/azuremonitor/#querying-azure-monitor-logs)',
+      InputParameters: [
+        {
+          Name: 'colName',
+          Type: 'System.String',
+          CslType: 'string',
+          CslDefaultValue: 'colName',
+        },
+        {
+          Name: '$myVar',
+          Type: 'System.String',
+          CslType: 'string',
+          CslDefaultValue: '$myVar',
+        },
+      ],
+      OutputColumns: [],
+    },
   };
 
   // Add template variables
--- a/src/components/QueryEditor/Suggestions.ts
+++ b/src/components/QueryEditor/Suggestions.ts
@@ -72,7 +72,7 @@ export function getFunctions(variables: VariableModel[]): Record<string, AdxFunc
         'If `$myVar` has the value `value1`,`value2`, it expands to: `colName in (value1,value2)`.' +
         'If using the `All` option, then check the `Include All Option` checkbox and in the `Custom all value` field type in the value `all`. If `$myVar` has value `all` then the macro will instead expand to `1 == 1`.' +
         'For template variables with a lot of options, this will increase the query performance by not building a large "where..in" clause.' +
-        '[Grafana docs](https://grafana.com/docs/grafana/latest/datasources/azuremonitor/#querying-azure-monitor-logs)',
+        '[Grafana docs](https://grafana.com/grafana/plugins/grafana-azure-data-explorer-datasource/)',
       InputParameters: [
         {
           Name: 'colName',
--- a/src/query_builder.test.ts
+++ b/src/query_builder.test.ts
@@ -1,9 +1,51 @@
 import interpolateKustoQuery from './query_builder';
+import { TemplateSrv } from '@grafana/runtime';
+import { escapeSpecial } from 'datasource';
 
 describe('QueryBuilder', () => {
+  const variableProps = (name: string, overrides?: any) => ({
+    name,
+    id: name,
+    multi: true,
+    includeAll: false,
+    type: 'custom',
+    ...overrides,
+  });
+  const variables = new Map<string, any>([
+    ['$__all', { ...variableProps('__all', { text: ['All'] }) }],
+    ['$test1', { ...variableProps('test1', { text: ['val1'] }) }],
+    ['$test2', { ...variableProps('test2', { text: ['val1', 'val2'] }) }],
+    ['$test3', { ...variableProps('test3', { text: ["'val1'"] }) }],
+    ['$test4', { ...variableProps('test4', { text: ["'val1'", "'val2'"] }) }],
+    ['$test5', { ...variableProps('test5', { text: ['test (with parentheses)', 'test without parentheses'] }) }],
+    ['$test6_a', { ...variableProps('test6_a', { text: ['test (with parentheses)', 'test without parentheses'] }) }],
+    ['$test6_b', { ...variableProps('test6_b', { text: ['test 2 (with parentheses)'] }) }],
+  ]);
+  const interpolateVariable = jest.fn().mockImplementation((value: any) => {
+    if (value.length === 1 && value[0] === 'All') return value;
+    else {
+      return value.map((val) => "'" + escapeSpecial(val) + "'").join(',');
+    }
+  });
+  const templateSrv: TemplateSrv = {
+    replace: jest.fn().mockImplementation((target: string, _scopedVars = undefined, format) => {
+      const split = target.split(',');
+      const variable = split[1].trim();
+      console.log(variable, variables.get(variable));
+      return `${split[0]}, ${format(variables.get(variable).text)}`;
+    }),
+    getVariables: jest.fn(),
+    containsTemplate: jest.fn(),
+    updateTimeRange: jest.fn(),
+  };
+
   describe('when $__contains and multi template variable has custom All value', () => {
     it('should generate a where..in clause', () => {
-      const query = interpolateKustoQuery('query=Tablename | where $__contains(col, all)');
+      const query = interpolateKustoQuery(
+        `query=Tablename | where $__contains(col, $__all)`,
+        templateSrv,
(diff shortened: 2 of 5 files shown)


andresmgot (commented):
Thanks for working on this! Regexes are tricky :sweat_smile:

src/query_builder.ts
```diff
@@ -4,7 +4,7 @@ export default function interpolateKustoQuery(query: string, scopedVars?: Scoped
   if (!query) {
     return '';
   }
-  const macroRegexp = /\$__([_a-zA-Z0-9]+)\(([^\)]*)\)/gi;
+  const macroRegexp = /\$__([_a-zA-Z0-9]+)\((.*)\)/gi;
```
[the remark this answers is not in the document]
aangelisc: Great spot! Regexes are definitely tricky 😜 I'll try your suggestion and maybe look into alternatives (I sometimes feel regex just adds additional complexity).

aangelisc:
I've updated the regex here, a new caveat being that this will only work with multiple contains separated by a '|'. I think this is a fair compromise otherwise the regex will get more complex with things like lookahead/behind. Wdyt @andresmgot?

andresmgot:
I was going to suggest to also cover the case in which the macros are separated by `or` or `and` but that won't work if the variable value contains those words. This solution won't work either if the value contains a `|` (e.g. `value | values`).

A process that should work in all cases (I believe) is:
 - Call `interpolateKustoQuery` without replacing variables
 - Once the regexp is applied, replace template variables (this would be within `getMultiContains` for example).
 - After replacing all the macros, call `templateSrv.replace` again with the whole query to replace the remaining template variables

That way we keep the regexp simple and we cover all the cases, wdyt?

aangelisc:
> I was going to suggest to also cover the case in which the macros are separated by `or` or `and` but that won't work if the variable value contains those words. This solution won't work either if the value contains a `|` (e.g. `value | values`).
> 
> A process that should work in all cases (I believe) is:
> 
> * Call `interpolateKustoQuery` without replacing variables
> * Once the regexp is applied, replace template variables (this would be within `getMultiContains` for example).
> * After replacing all the macros, call `templateSrv.replace` again with the whole query to replace the remaining template variables
> 
> That way we keep the regexp simple and we cover all the cases, wdyt?

This is a much better process! I'll implement this. Great thinking 😊

src/query_builder.ts
```diff
@@ -1,6 +1,12 @@
 import { ScopedVars } from '@grafana/data';
+import { TemplateSrv } from '@grafana/runtime';
 
-export default function interpolateKustoQuery(query: string, scopedVars?: ScopedVars): string {
+export default function interpolateKustoQuery(
+  query: string,
+  templateSrv: TemplateSrv,
+  interpolateVariable: (value: any, variable: any) => string | number,
```
andresmgot, line 7: nit: you can save one parameter if you pass a function: `replace: (val: string) => string` and call it with: `(val: string) => templateSrv.replace(p2, scopedVars, interpolateVariable)`

src/query_builder.ts
```diff
@@ -9,7 +15,8 @@ export default function interpolateKustoQuery(query: string, scopedVars?: Scoped
 
   query = query.replace(macroRegexp, (match, p1, p2) => {
     if (p1 === 'contains') {
-      return getMultiContains(p2);
+      const replaced = templateSrv.replace(p2, scopedVars, interpolateVariable);
+      return getMultiContains(replaced);
     }
     return match;
   });
```
andresmgot, line 22: You also need to call `replace` when evaluating `$__escapeMulti` below since that can be called with a template variable as an argument.

andresmgot (approved):
Nice! Thanks for also adding the `Suggestion` :+1: One comment left but once that is done you can merge this

src/query_builder.ts
aangelisc: I've made some changes to this (it was a bit bigger than just putting `replace` in there). Lmk what you think 😊

src/components/QueryEditor/Suggestions.ts
```diff
@@ -63,6 +63,32 @@ export function getFunctions(variables: VariableModel[]): Record<string, AdxFunc
...
+    $__contains: {
+      Name: '$__contains',
+      Body: `{ colName in ('value1','value2') }`,
+      FunctionKind: 'Unknown',
+      DocString:
+        '##### Used with multi-value template variables.\n\n' +
+        'If `$myVar` has the value `value1`,`value2`, it expands to: `colName in (value1,value2)`.' +
+        'If using the `All` option, then check the `Include All Option` checkbox and in the `Custom all value` field type in the value `all`. If `$myVar` has value `all` then the macro will instead expand to `1 == 1`.' +
+        'For template variables with a lot of options, this will increase the query performance by not building a large "where..in" clause.' +
+        '[Grafana docs](https://grafana.com/docs/grafana/latest/datasources/azuremonitor/#querying-azure-monitor-logs)',
```
andresmgot, line 75: this link is pointing to azure monitor

src/query_builder.test.ts
```diff
@@ -1,70 +1,172 @@
...
+    name,
+    id: name,
+    multi: true,
+    includeAll: false,
+    type: 'custom',
+    ...overrides,
+  });
+  const variables = new Map<string, any>([
+    ['$__all', { ...variableProps('__all', { text: ['All'] }) }],
+    ['$test1', { ...variableProps('test1', { text: ['val1'] }) }],
```
andresmgot, line 16: nit: rather than numbers, it would be better to have a description of the variable, for example `$test_one_var` `$test_two_vars`, `$test_quoted_var`...

src/query_builder.test.ts
```diff
@@ -1,70 +1,172 @@
...
+    describe('when using $__escape and multi template variable has one selected value', () => {
+      it('should replace $__escape(val) with KQL style escaped string', () => {
+        templateSrv.replace = escapeReplace;
+        const query = interpolateKustoQuery(`$__escapeMulti($test7)`, (val: string) =>
+          templateSrv.replace(val, undefined, interpolateVariable)
+        );
+        expect(query).toContain(`@'\\grafana-vmNetwork(eth0)Total Bytes Received'`);
+      });
+    });
+    describe('when using $__escape and multi template variable has multiple selected values', () => {
```
andresmgot, line 142: nit: (style) leave one empty line between tests

src/query_builder.ts
```diff
@@ -1,15 +1,20 @@
...
+export default function interpolateKustoQuery(
+  query: string,
+  replace: (val: string) => string,
+  scopedVars?: ScopedVars
+): string {
   if (!query) {
     return '';
   }
-  const macroRegexp = /\$__([_a-zA-Z0-9]+)\(([^\)]*)\)/gi;
+  const macroRegexp = /\$__([_a-zA-Z0-9]+)\(([^\|]*[^\)])\)/gi;
```
andresmgot, line 11: why not just leaving it as the original regex?

src/query_builder.ts
```diff
@@ -22,7 +27,7 @@ export default function interpolateKustoQuery(query: string, scopedVars?: Scoped
     return values?.value ?? match;
   });
 
-  return query.replace(/\$__escapeMulti\(('[^]*')\)/gi, (match, p1) => escape(p1));
+  return query.replace(/\$__escapeMulti\(([^\|]*[^\)])\)/gi, (match, p1) => escape(replace(p1)));
```
andresmgot, line 30: same, why did you need to change the regex?

andresmgot (approved):
Good job here! I just have `nit` comments and questions

src/query_builder.ts
aangelisc: The original regex does not work here as values are no longer quoted because we aren't doing the replace and interpolation before executing this function. However, I think the original macro regexp is the better one now: `\$__([_a-zA-Z0-9]+)\(([^\)]*)\)` wdyt? This could be used above also.

andresmgot: if that regexp works for this case yeah, better to call this `escape` in the first regex evaluation, we can skip this special case

src/query_builder.ts
```diff
@@ -9,7 +13,12 @@ export default function interpolateKustoQuery(query: string, scopedVars?: Scoped
 
   query = query.replace(macroRegexp, (match, p1, p2) => {
```
andresmgot, line 14: one last nit :sweat_smile: you can do:

```ts
return replace(query.replace ...
```

that way, functions calling `interpolateKustoQuery` do not need to call `templateSrv.replace` again, it's automatically done here.

andresmgot (approved):

src/query_builder.ts
aangelisc: Do you mean replace the final `return` on line 34 with `return replace(query, ...)`? This would then mean we could remove lines 97 and 233 in `datasource.ts`

andresmgot: yes, and that you don't need to redefine `query = query`, simply return the value

aangelisc: I think we can't return immediately at this point as we still have the interval regex to run and replace any interval variables. Unless I'm missing something?

src/query_builder.ts
andresmgot: you are right, I was looking at the github view in which that piece of code was collapsed so I didn't see it :sweat_smile:

andresmgot (approved):
:+1:

src/query_builder.ts
andresmgot: (for some reason this message was not sent)

You are right!, I was looking at the github view that collapsed the other replacing :)

PR #483, opened 2022-10-03
fix: autocomplete now works with dynamic values

### LINKED ISSUE #453: Failed to autocomplete dynamic values with arrays
**What happened**:

When a column is dynamic and contains an array, the query to get suggestions is wrong so the result is empty:

Screenshot from 2022-08-25 09-45-48

To get suggestions about possible values

**Screenshots**

The[ `autoComplete`](https://github.com/grafana/azure-data-explorer-datasource/blob/main/src/datasource.ts#L292) query is wrong. It's executing:

```
DestinyActivities
| where $__timeFilter(TimeGenerated)
| where isnotempty(Modes["`indexer`"])
| take 50000
| distinct tostring(Modes["`indexer`"])
| take 251
```

While the correct query should be:

```
DestinyActivities
| where $__timeFilter(TimeGenerated)
| mv-expand array_1 = Modes
| where isnotempty(array_1)
| take 50000
| distinct tostring(array_1)
| take 251
```

This has been already solved in the query builder but the extra parsing is not happening in for auto complete queries.

**Environment**:

- Grafana version:
- Plugin version: 4.1.5
- OS Grafana is installed on:
- User OS & Browser:
- Others:

asimpson:
This should work for both old and new editors.

fixed-auto

Closes #453

### DIFF
--- a/src/KustoExpressionParser.ts
+++ b/src/KustoExpressionParser.ts
@@ -43,17 +43,25 @@ export class KustoExpressionParser {
     };
 
     const parts: string[] = [];
+    const isDynamic = query.search.property.name.includes(DYNAMIC_TYPE_ARRAY_DELIMITER);
+    const name: string = this.addExpanPartsdIfNeeded(query.search.property.name, parts);
     this.appendProperty(context, query.expression.from, parts);
     this.appendTimeFilter(context, undefined, parts, tableSchema);
 
-    if (query.index) {
+    //query.index is used by the legacy query editor
+    if (query.index && !isDynamic) {
       const where = replaceByIndex(query.index, query.expression.where, query.search);
       this.appendWhere(context, where, parts, 'where');
     }
-    const column = query.search.property.name;
+
+    if (isDynamic) {
+      parts.push(`mv-expand ${name} = ${parts[0]}`);
+      parts.push(`where ${query.search.operator.name}(${name})`);
+      parts.shift();
+    }
 
     parts.push('take 50000');
-    parts.push(`distinct ${context.castIfDynamic(column)}`);
+    parts.push(`distinct ${context.castIfDynamic(name, query.search.property.name)}`);
     parts.push('take 251');
 
     return parts.join('\n| ');
--- a/src/KustoExpressionParser.ts
+++ b/src/KustoExpressionParser.ts
@@ -57,7 +57,7 @@ export class KustoExpressionParser {
     }
 
     //query.index is used by the legacy query editor
-    //isDynamicArray keeps the old behavior of the legacy query editor for non dynamic arrays
+    //!isDynamicArray is needed for when the legacy query editor is used with a non-dynamic value and we need to replace by index
     if (query.index && !isDynamicArray) {
       const where = replaceByIndex(query.index, query.expression.where, query.search);
       this.appendWhere(context, where, parts, 'where');
--- a/src/KustoExpressionParser.ts
+++ b/src/KustoExpressionParser.ts
@@ -44,7 +44,6 @@ export class KustoExpressionParser {
 
     const parts: string[] = [];
     const expandParts: string[] = [];
-    const isDynamicArray = query.search.property.name.includes(DYNAMIC_TYPE_ARRAY_DELIMITER);
     const name: string = this.addExpanPartsdIfNeeded(query.search.property.name, expandParts);
     const column = context.castIfDynamic(name, query.search.property.name);
     this.appendProperty(context, query.expression.from, parts);
@@ -57,8 +56,7 @@ export class KustoExpressionParser {
     }
 
     //query.index is used by the legacy query editor
-    //!isDynamicArray is needed for when the legacy query editor is used with a non-dynamic value and we need to replace by index
-    if (query.index && !isDynamicArray) {
+    if (query.index) {
       const where = replaceByIndex(query.index, query.expression.where, query.search);
       this.appendWhere(context, where, parts, 'where');
     } else {
--- a/src/KustoExpressionParser.ts
+++ b/src/KustoExpressionParser.ts
@@ -43,25 +43,30 @@ export class KustoExpressionParser {
     };
 
     const parts: string[] = [];
-    const isDynamic = query.search.property.name.includes(DYNAMIC_TYPE_ARRAY_DELIMITER);
-    const name: string = this.addExpanPartsdIfNeeded(query.search.property.name, parts);
+    const expandParts: string[] = [];
+    const isDynamicArray = query.search.property.name.includes(DYNAMIC_TYPE_ARRAY_DELIMITER);
+    const name: string = this.addExpanPartsdIfNeeded(query.search.property.name, expandParts);
+    const column = context.castIfDynamic(name, query.search.property.name);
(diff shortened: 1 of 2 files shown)


src/KustoExpressionParser.ts
```diff
@@ -43,17 +43,25 @@ export class KustoExpressionParser {
     };
 
     const parts: string[] = [];
+    const isDynamic = query.search.property.name.includes(DYNAMIC_TYPE_ARRAY_DELIMITER);
```
asimpson, line 46: I know the schema has a property `isDynamic` that we could check instead but checking this way avoids having to filter through that list.

src/KustoExpressionParser.ts
```diff
@@ -43,17 +43,25 @@ export class KustoExpressionParser {
...
+    if (query.index && !isDynamic) {
       const where = replaceByIndex(query.index, query.expression.where, query.search);
       this.appendWhere(context, where, parts, 'where');
     }
-    const column = query.search.property.name;
+
+    if (isDynamic) {
+      parts.push(`mv-expand ${name} = ${parts[0]}`);
+      parts.push(`where ${query.search.operator.name}(${name})`);
+      parts.shift();
```
asimpson, line 60: Needed since the top of `parts` at this point will have the name of array, e.g. `Modes`.

src/KustoExpressionParser.ts
yaelleC: the delimiter is `'["`indexer`"]'` so safe to say it will not be included in the property name by chance 👍

yaelleC (approved):
LGTM 🚀

src/KustoExpressionParser.ts
andresmgot: nit: I would call this `isDynamicArray` to be more accurate

src/KustoExpressionParser.ts
andresmgot: you need to do this because `addExpanPartsdIfNeeded` is meant to be called with an specific array for "parts to expand". If you define that array, you can use here the predefined function to add `mv-expand`.

Also, you can use the predefined function to `appendWhere` to do the condition. The only thing that you need to modify is the `search` query since it's still pointing to the "unexpanded" column name.

src/KustoExpressionParser.ts
```diff
@@ -43,17 +43,25 @@ export class KustoExpressionParser {
...
 
     const parts: string[] = [];
+    const isDynamic = query.search.property.name.includes(DYNAMIC_TYPE_ARRAY_DELIMITER);
+    const name: string = this.addExpanPartsdIfNeeded(query.search.property.name, parts);
     this.appendProperty(context, query.expression.from, parts);
     this.appendTimeFilter(context, undefined, parts, tableSchema);
 
-    if (query.index) {
+    //query.index is used by the legacy query editor
+    if (query.index && !isDynamic) {
```
andresmgot, line 52: I just noticed the current code has a bug, if there is no index, we are not appending the condition `isnotempty(col)`. If query.index is not defined, it should still append that.

andresmgot (commented):
Nice job navigating all the complexity here! It's difficult to untangle how all these functions work. My suggestion here is to use the predefined functions that there are meant for this stuff. Basically with the suggestions I give in-line, the code can be something like:

```ts
    const parts: string[] = [];
    const expandParts: string[] = [];
    const name: string = this.addExpanPartsdIfNeeded(query.search.property.name, expandParts);
    this.appendProperty(context, query.expression.from, parts);
    this.appendTimeFilter(context, undefined, parts, tableSchema);
    this.appendMvExpand(expandParts, parts);

    if (expandParts.length) {
      // Replace the column name in the search expression with the "expanded" column name
      query.search.property.name = name;
    }

    //query.index is used by the legacy query editor
    if (query.index) {
      const where = replaceByIndex(query.index, query.expression.where, query.search);
      this.appendWhere(context, where, parts, 'where');
    } else {
      this.appendWhere(context, query.search, parts, 'where');
    }
```

src/KustoExpressionParser.ts
asimpson: This get's appended [here](https://github.com/grafana/azure-data-explorer-datasource/blob/df669efb46ba75232491ae829ffe0ac0249d7b5e/src/KustoExpressionParser.ts#L283) if I'm understanding this correctly. Should already be good to go.

asimpson:
@andresmgot Thanks for the push to utilize the existing functions more! Passing `query.search` to `appendWhere` was the unlock for me 🏆 

I kept `isDynamicArray` around since we need it for dynamic arrays in the old editor. Other than that I incorporated your suggestions and this should be good. I'll merge it first thing tomorrow barring any additional feedback. 🙇

src/KustoExpressionParser.ts
```diff
@@ -43,17 +43,30 @@ export class KustoExpressionParser {
...
     const parts: string[] = [];
+    const expandParts: string[] = [];
+    const isDynamicArray = query.search.property.name.includes(DYNAMIC_TYPE_ARRAY_DELIMITER);
+    const name: string = this.addExpanPartsdIfNeeded(query.search.property.name, expandParts);
+    const column = context.castIfDynamic(name, query.search.property.name);
     this.appendProperty(context, query.expression.from, parts);
     this.appendTimeFilter(context, undefined, parts, tableSchema);
+    this.appendMvExpand(expandParts, parts);
+
+    if (expandParts.length) {
```
andresmgot, line 54: nit: this can be `if (isDynamicArray)` (you may want to move it under the other else?

src/KustoExpressionParser.ts
```diff
@@ -43,17 +43,30 @@ export class KustoExpressionParser {
...
+      // Replace the column name in the search expression with the "expanded" column name
+      query.search.property.name = name;
+    }
 
-    if (query.index) {
+    //query.index is used by the legacy query editor
+    //isDynamicArray keeps the old behavior of the legacy query editor for non dynamic arrays
```
andresmgot, line 60: I don't really understand this comment? I guess that for arrays, you are not attempting to replace the unfinished `where` condition (the one we are trying to get suggestions for) from the list of filters, but I am not sure why?

asimpson: Yeah sorry it wasn't clear. I updated the comment so hopefully it's more clear? Essentially we're checking for the case of: 

1. the legacy query editor
2. a non-dynamic value, e.g. "ActivityName" 

If those two conditions are met we need to replace by index for autocomplete to continue to work correctly. Here's a screenshot of the case I'm talking about to hopefully clear up what I mean (words are hard today :see_no_evil: ).
image

andresmgot: yeah, my question is, why is it different for dynamic arrays? In your example, if you say `ActivityName == Football` and add a new filter `Modes[] == `, then as suggestions you will get all the possible values, including for rows in which `ActivityName` is not `Footbal` (which is the case anyway for the new editor because we are not passing down that index yet). It's not a blocking thing, but just curious why it was that way (in case we add the index in the new editor).

andresmgot: cool! Thanks for checking!

PR #484, opened 2022-10-05
chore: prepare 4.1.7 release

asimpson:


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Change Log
 
+## [4.1.7]
+
+- Fix state changes in useEffect
```
andresmgot, line 5: these notes are for users so rather than the internal change description we better show the user-facing change. For example, here I believe it would be "Fix crash when creating alerts" (or something like that)

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
...
+- Pre-select columns for new queries
```
andresmgot, line 6: this is not visible so you can skip this point

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
...
+- Update flow for variable interpolation
```
andresmgot, line 7: same here (maybe we should use more user-friendly PR titles :P). I believe this would be "Fix template variables for values containing parentheses"

andresmgot (commented):
Thanks for preparing the release! I have some suggestions for the changelog

Issue #485, opened 2022-10-06
Add feature tracking


andresmgot:
**What happened**:

There are multiple features in ADX that due to its complexity or the very specific use cases they cover, are probably not used. The goal of this task is to add some feature tracking so we can get more information about how much these features are used. If they are not, we can remove them from the code base, if they are we can improve its UX so they are easier to use.

This is the list of things we can track:

Data source configuration:
- On-Behalf-Of authentication: This is currently a beta feature, quite difficult to setup and recently added.
- Query Timeout: This is a differen timeout than Grafana's data proxy. It's not useful if the data proxy is not configured with at least the same value.
- Dynamic caching and max cache age. This is quite easy to use and pretty valuable, but disabled by default. It would be interesting to see its usage so maybe we can default it to `true`.
- Data consistency: It's not clear what this does so I'd assume not that many people use it.
- Default editor value: Is there a point of using "Raw" by default?
- Managed schema: Seems a bit difficult to use, allowing just certain schema values.

Query editor:

- Format: I assume a big percentage of queries will use the "Time series" format, since it was the default but we lately changed it to "Table". Also there is a third format "ADX Time series" which we don't know how much it's used.
- Builder vs KQL editor: How much is the raw (KQL) editor used? Should we invest more on it?

andresmgot:
Question here! Maybe @yaelleC you have an opinion:

For data source features (e.g. usage of on-behalf-of), we have two options:

 - Track 1 event per data source load. This will give us data only about the data sources (e.g. 5% of data sources enable On-Behalf-Of authentication)
 - Track 1 event per dashboard load. Similar to what we do with query features (e.g. query type). In this case, the focus will be on queries. For example: 5% queries uses a data source with On-Behalf-Of authentication.

I believe the second is more valuable since it's more relevant a data source used in 1k queries than one used just for one but maybe I am missing something.

Also, there is the issue of retrieving the data source information on the dashboard load, which should be possible but haven't tried yet.

yaelleC:
Another option would be to add the event on `save-and-test` button. E.g 5% of datasources were set up using xyz feature (not sure how that would differ from your first option).

But I agree with you, having the data on dashboard loads to focus on volume might be more interesting, but I'm not sure either how possible it would be to get some feature information on dashboard load.
Which brings to mind a newbie / silly question: How is authentication handled on queries? Is it a one of (for the dashboard / datasource loaded in state)? or is it done for every query?

[...]

PR #487, opened 2022-10-10
Add initial feature tracking

andresmgot:
Query features from: #485

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -21,7 +21,7 @@ getAppEvents().subscribe<DashboardLoadedEvent<KustoQuery>>(
   DashboardLoadedEvent,
   ({ payload: { dashboardId, orgId, grafanaVersion, queries } }) => {
     const adxQueries = queries[pluginJson.id]?.filter((q) => !q.hide);
-    if (!adxQueries?.length) {
+    if (adxQueries && adxQueries.length > 0) {
       return;
     }
     const counters = {
@@ -46,14 +46,12 @@ getAppEvents().subscribe<DashboardLoadedEvent<KustoQuery>>(
       query.rawMode ? counters.raw_queries++ : counters.query_builder_queries++;
     });
 
-    if (adxQueries && adxQueries.length > 0) {
-      trackADXMonitorDashboardLoaded({
-        adx_plugin_version: pluginJson.info.version,
-        grafana_version: grafanaVersion,
-        dashboard_id: dashboardId,
-        org_id: orgId,
-        ...counters,
-      });
-    }
+    trackADXMonitorDashboardLoaded({
+      adx_plugin_version: pluginJson.info.version,
+      grafana_version: grafanaVersion,
+      dashboard_id: dashboardId,
+      org_id: orgId,
+      ...counters,
+    });
   }
 );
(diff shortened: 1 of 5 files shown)


src/module.ts
```diff
@@ -1,14 +1,59 @@
...
   .setConfigEditor(ConfigEditor)
   .setQueryEditorHelp(EditorHelp)
   .setQueryEditor(QueryEditor);
+
+// Track dashboard loads to RudderStack
+getAppEvents().subscribe<DashboardLoadedEvent<KustoQuery>>(
+  DashboardLoadedEvent,
+  ({ payload: { dashboardId, orgId, grafanaVersion, queries } }) => {
+    const adxQueries = queries[pluginJson.id]?.filter((q) => !q.hide);
+    if (!adxQueries?.length) {
```
aangelisc, line 24:
```suggestion
    if (adxQueries && adxQueries.length > 0) {
```

As the condition around `trackADXMonitorDashboardLoaded` can be removed

src/tracking.ts
```diff
@@ -0,0 +1,35 @@
...
+ * Dashboard: https://ops.grafana.net/d/Ad0pti0N/data-sources-adoption-tracking?orgId=1
+ * Changelog:
+ * - 4.1.7: Initial version
+ */
+export const trackADXMonitorDashboardLoaded = (props: ADXMonitorDashboardLoadedProps) => {
+  reportInteraction('grafana_ds_adx_dashboard_loaded', props);
+};
+
+export type ADXMonitorDashboardLoadedProps = {
+  adx_plugin_version?: string;
```
yaelleC, line 21: 🚀 nice!

src/module.ts
```diff
@@ -1,14 +1,57 @@
...
 )
   .setConfigEditor(ConfigEditor)
   .setQueryEditorHelp(EditorHelp)
   .setQueryEditor(QueryEditor);
+
+// Track dashboard loads to RudderStack
+getAppEvents().subscribe<DashboardLoadedEvent<KustoQuery>>(
+  DashboardLoadedEvent,
+  ({ payload: { dashboardId, orgId, grafanaVersion, queries } }) => {
+    const adxQueries = queries[pluginJson.id]?.filter((q) => !q.hide);
```
yaelleC, line 23: Keeping only the visible queries: sounds good!

yaelleC (approved):
LGTM
Just curious about the difference between `FormatOptions.adxTimeSeries` and `FormatOptions.TimeSeries` I thought we only had two possible format options ?

andresmgot:
> Just curious about the difference between `FormatOptions.adxTimeSeries` and `FormatOptions.TimeSeries` I thought we only had two possible format options ?

It's kind of an advanced feature. That format is only available when using the KQL editor, since it requires to use the `make-series` Kusto operator, which is not something that you can add using the query builder. It's documented in our [docs](https://github.com/grafana/azure-data-explorer-datasource#adx-time-series-queries).

[Screencast from 10-10-22 14:48:22.webm](https://user-images.githubusercontent.com/4025665/194870079-cb011762-39a1-489e-a280-184b2671561c.webm)

PR #488, opened 2022-10-11
Add data source to feature tracking

### LINKED ISSUE #485: Add feature tracking

andresmgot:
Closes #485

Add the rest of features that we can to track based on its relevant data source configuration. There is no issue getting the data source settings at that time (it's even synchronous so that simplifies things).

The code that counted queries was getting bulky so I moved to its own function and added some tests.

### DIFF
--- a/src/tracking.ts
+++ b/src/tracking.ts
@@ -1,4 +1,6 @@
-import { reportInteraction } from '@grafana/runtime';
+import { DataSourceInstanceSettings } from '@grafana/data';
+import { DataSourceSrv, reportInteraction } from '@grafana/runtime';
+import { AdxDataSourceOptions, EditorMode, FormatOptions, KustoQuery } from 'types';
 
 /**
  * Loaded the first time a dashboard containing ADX queries is loaded (not on every render)
@@ -13,15 +15,12 @@ import { reportInteraction } from '@grafana/runtime';
  * Changelog:
  * - 4.1.7: Initial version
  */
-export const trackADXMonitorDashboardLoaded = (props: ADXMonitorDashboardLoadedProps) => {
+export const trackADXMonitorDashboardLoaded = (props: ADXDashboardLoadedProps) => {
+  console.log('ADXMonitorDashboardLoaded', props);
   reportInteraction('grafana_ds_adx_dashboard_loaded', props);
 };
 
-export type ADXMonitorDashboardLoadedProps = {
-  adx_plugin_version?: string;
-  grafana_version?: string;
-  dashboard_id: string;
-  org_id?: number;
+export type ADXCounters = {
   /** number of queries using the "Table" format  */
   table_queries: number;
   /** number of queries using the "Time Series" format  */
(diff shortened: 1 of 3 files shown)


aangelisc (approved):
LGTM, I'd love to think of a way of shortening the settings checks in `tracking.ts` but still maintaining readability (which rules out ternaries).

src/tracking.ts
```diff
@@ -32,4 +30,85 @@ export type ADXMonitorDashboardLoadedProps = {
...
+  /** number of queries using a timeout different than the default */
+  queries_with_custom_timeout: number;
+  /** number of queries using ADX dynamic caching */
+  dynamic_caching_queries: number;
+  /** number of queries using weak data consistency (not default) */
+  weak_data_consistency_queries: number;
+  /** number of queries using the Kusto editor by default (not default) */
+  queries_with_default_raw_editor: number;
+  /** number of queries using a custom schema mapping */
+  queries_with_managed_schema: number;
```
yaelleC, line 44: nice, they all seem interesting to know about 👍

src/tracking.ts
```diff
@@ -13,15 +15,11 @@ import { reportInteraction } from '@grafana/runtime';
  * Changelog:
  * - 4.1.7: Initial version
  */
-export const trackADXMonitorDashboardLoaded = (props: ADXMonitorDashboardLoadedProps) => {
+export const trackADXMonitorDashboardLoaded = (props: ADXDashboardLoadedProps) => {
```
yaelleC, line 18: Oops 😄

yaelleC (approved):
LGTM 🚀 

I guess the only thing that comes to mind is why choose to have only one prop for things like OBO, Dynamic caching, default to raw... (assuming total - prop = number of opposite thing) and to have both `query_builder_queries` and `raw_queries` even though they are mutually exclusive?

andresmgot:
> I guess the only thing that comes to mind is why choose to have only one prop for things like OBO, Dynamic caching, default to raw... (assuming total - prop = number of opposite thing) and to have both `query_builder_queries` and `raw_queries` even though they are mutually exclusive?

Kind of a small optimization, we need at least one mutual exclusive set of properties ( `query_builder_queries` and `raw_queries`) to get that total. Once we have it, it's redundant to have the other option for the rest of properties (also it's kind of trivial to get the meaning e.g. `queries_with_custom_timeout` vs `queries_without_custom_timeout`).

I could remove one of the format options and use the total but since those are three options, I think it's more readable to just leave those three.

yaelleC:
> we need at least one mutual exclusive set of properties

Technically we would still get an event fired, even if all the counters are at `0` so we could count events.
But for the sake of having it easy when we visualise the data  I am happy merging the code as is.  👍

asimpson (approved):
Nice!

PR #493, opened 2022-10-17
Replace check with a comment for Levitate result

andresmgot:
- Use GH action rather than calling `npx` directly
- Set Levitate as a comment rather than a check
- Upgrade @grafana/* packages to make Levitate happy (unless e2e packages since I found a bug with those)
- Adapt code to the new versions


aangelisc (approved):
Thanks for fixing this 👍

Issue #405, opened 2022-06-10
Macros displayed as KQL functions


andresmgot:
**What happened**:

One of the actions we did while working on the raw query editor revamp was to define Grafana macros as “functions'' so they can be autocompleted. These macros are now being displayed when configuring schema mappings but they are not real database functions so they shouldn’t. It only happens when the editor is loaded before.

**What you expected to happen**:

Show only real functions (if any)

**How to reproduce it (as minimally and precisely as possible)**:

1. Go to the query editor and load the raw editor.
2. Go to the configuration page.
3. Enable schema mappings.
4. See that macros are listed there.

**Screenshots**

**Anything else we need to know?**:

**Environment**:

- Grafana version:
- Plugin version: 4.1.0
- OS Grafana is installed on:
- User OS & Browser:
- Others:

PR #494, opened 2022-10-18
Avoid modifying the database schema

andresmgot:
The raw query editor was modifying the `schema` object stored, causing the data source configuration page to show some wrong data.

Fixes https://github.com/grafana/azure-data-explorer-datasource/issues/405

Note: This only fixes the new query editor, not the legacy one.


asimpson (approved):
:boom:

PR #498, opened 2022-10-20
Refactor: Use version from plugin meta info

andresmgot:
The version that was being pass down was the string `%VERSION%` rather than the real version of the plugin.


aangelisc (approved):
👍

PR #506, opened 2022-11-01
Fix: Invalid code editor loaded for Grafana versions that don't follow semantic versioning

### LINKED ISSUE #505: Invalid code editor loaded for Grafana versions that don't follow semantic versioning
If the version of Grafana isn't a valid semantic version then the code editor for Grafana pre-8.5 will be loaded. This causes issues with other panels and leads to Grafana crashing.

aangelisc:
Will correctly handle security releases by coercing the version into a value that follows semver rules.

Fixes #505

### DIFF
--- a/src/components/LegacyQueryEditor/RawQueryEditor.tsx
+++ b/src/components/LegacyQueryEditor/RawQueryEditor.tsx
@@ -6,7 +6,7 @@ import { QueryEditorResultFormat, selectResultFormat } from './QueryEditorResult
 import { AdxDataSource } from 'datasource';
 import { KustoMonacoEditor } from 'monaco/KustoMonacoEditor';
 import React, { useEffect, useState } from 'react';
-import { gte, valid } from 'semver';
+import { gte, valid, coerce } from 'semver';
 import { selectors } from 'test/selectors';
 import { AdxDataSourceOptions, AdxSchema, KustoQuery } from 'types';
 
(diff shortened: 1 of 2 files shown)


src/components/LegacyQueryEditor/RawQueryEditor.tsx
```diff
@@ -41,6 +41,15 @@ interface Worker {
 // that includes fixes required for auto-completion to work.
 // Remove this code once Grafana 8.5 is the minimal version supported
 function gtGrafana8_5() {
+  const version = config.buildInfo.version
+  const isValid = valid(config.buildInfo.version)
```
asimpson, line 45: Nit: could just pass `version` to `valid` no?

src/components/LegacyQueryEditor/RawQueryEditor.tsx
```diff
@@ -41,6 +41,15 @@ interface Worker {
...
+  // Assume that a security release will be of the form 'w.x.y.z' - Pre releases of the form 'w.x.y-pre.z' are already valid
+  const isSecurityRelease = version.split('.').length > 3
+  if (!isValid && isSecurityRelease) {
+    // Coerce will drop the z in 'w.x.y.z' leaving just the major-minor-patch version
+    const coercedValue = coerce(version)
+    return coercedValue ? gte(coercedValue, '8.5.0') : null;
+  }
   return valid(config.buildInfo.version) && gte(config.buildInfo.version, '8.5.0');
```
asimpson, line 53: Nit: could return `isValid` here instead of re-running `valid`. Could also pass `version` instead of the full object address again.

asimpson (approved):
🚀 nice work!

✅ tests pass 
✅ solves the issue

I left two nit comments but this works for me!

andresmgot (approved):

PR #519, opened 2022-11-22
Correctly validate query property

### LINKED ISSUE #514: Potentially undefined prop leading to error unmounting ADX query editor
**What happened**:
Switching from the ADX KQL editor to another editor (e.g. Cloudwatch, Azure Monitor etc) leads to an error (as shown below). This is due to the `query.expression` property missing which causes errors when unmounting.

image

Switching to another editor should not cause an error.

aangelisc:
Checks if `expression` is undefined before attempting to access `from`

Fixes #514


andresmgot (approved):
:+1: thanks

PR #521, opened 2022-11-23
Add Grafana toolkit

aangelisc:
Switching to the Grafana toolkit for plugin development requires the package to be included for the scripts to be run.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "@grafana/e2e": "9.1.2",
     "@grafana/e2e-selectors": "9.1.2",
     "@grafana/eslint-config": "^2.5.0",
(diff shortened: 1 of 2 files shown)


andresmgot:
which scripts? I think that the goal was to remove the grafana/toolkit in favor of the `create-plugin` tool

andresmgot:
actually you may be referring to the `watch` script, I thought I removed it but it's still there. We should use `yarn dev` now.

aangelisc:
> actually you may be referring to the `watch` script, I thought I removed it but it's still there. We should use `yarn dev` now.

Ah okay, I will update this PR to remove the `watch` and `sign` scripts instead!

Issue #509, opened 2022-11-04
All requests should use user identity auth if user identity configured


kostrse:
When user configures OBO authentication then some requests authenticated using the user identity of OBO and some continue to use service identity directly.

Properly using user identity:

* **QueryData** ([here](https://github.com/grafana/azure-data-explorer-datasource/blob/e25f75a831651e5825bef183496d778c87624f50/pkg/azuredx/datasource.go#L82))

Methods using service identity directly (need to be fixed):

* CheckHealth([here](https://github.com/grafana/azure-data-explorer-datasource/blob/e25f75a831651e5825bef183496d778c87624f50/pkg/azuredx/datasource.go#L96))
* CallResource **getDatabases** ([here](https://github.com/grafana/azure-data-explorer-datasource/blob/fb3895246c4f2c4e8d2001edaa44e57a06fca8d6/pkg/azuredx/resource_handler.go#L57))
* CallResource **getSchema** ([here](https://github.com/grafana/azure-data-explorer-datasource/blob/fb3895246c4f2c4e8d2001edaa44e57a06fca8d6/pkg/azuredx/resource_handler.go#L28))

Regardless of type of the request, authentication method should be consistent with user's configuration.

kostrse:
This PR #525 isolates the issue to `CheckHealth` ([here](https://github.com/grafana/azure-data-explorer-datasource/blob/09ce9e1aa33cee133f80ae40d9eee4915c39d1a1/pkg/azuredx/client/client.go#L46)).

Short term solution for `CheckHealth` would be to temporary remove server-side implementation and implement it via `testDatasource()` on frontend which would make an normal query request.

Ideal solution would be to be to include `IdToken` to `req.PluginContext.User` ([here](https://github.com/grafana/grafana-plugin-sdk-go/blob/2edccc9894df92d77ba3b1a80ed982b82e7ba437/backend/common.go#L14)), so ID token could be accessed from the `User` struct regardless of which backend func was called (query req, resource call, etc.).

PR #525, opened 2022-12-13
CurrentUserContext to carry current user identity to the authentication layer

kostrse:
The final goal is to completely separate authentication logic from the business layer. The token request logic should not live  alongside with normal query code. This will allow to abstract out authentication from the core plugin implementation and eventually move authentication to the Grafana Azure SDK.

This is intermediate PR which introduces `CurrentUserContext` context abstraction which is supposed to carry current user identity though context down to the transport layer.

ADX client extracts the `CurrentUserContext` and uses it for authentication regardless of where the context was originated.

Helper methods `adxusercontext.FromQueryReq` and `adxusercontext.FromResourceReq` provided to create user context for query ([here](https://github.com/grafana/azure-data-explorer-datasource/blob/09ce9e1aa33cee133f80ae40d9eee4915c39d1a1/pkg/azuredx/datasource.go#L90)) and resource ([here](https://github.com/grafana/azure-data-explorer-datasource/blob/09ce9e1aa33cee133f80ae40d9eee4915c39d1a1/pkg/azuredx/datasource.go#L103)) requests. After the context was created, it can be passed to the rest of the business logic, down to ADX client without having to have any handling of access token or authorization header.

### Problem with heath check

Due to current design of Plugin SDK, `CheckHealth` doesn't carry ID token (see [here](https://github.com/grafana/azure-data-explorer-datasource/blob/09ce9e1aa33cee133f80ae40d9eee4915c39d1a1/pkg/azuredx/adxauth/adxusercontext/from_req.go#L52)), so this PR has a workaround specifically for `TestRequest` to use service identity via `GetServiceAccessToken(ctx)` for datasource testing ([here](https://github.com/grafana/azure-data-explorer-datasource/blob/09ce9e1aa33cee133f80ae40d9eee4915c39d1a1/pkg/azuredx/client/client.go#L46)), rather than using the proper user identity.

This is incorrect behavior which already existed before this PR. This PR just isolated it to `CheckHealth`/`TestRequest` so the issue becomes more straightforward to address.

* #509
* https://github.com/grafana/grafana-plugin-sdk-go/issues/579

### DIFF
--- a/pkg/azuredx/adxauth/adxusercontext/from_req.go
+++ b/pkg/azuredx/adxauth/adxusercontext/from_req.go
@@ -46,7 +46,8 @@ func FromHealthCheckReq(ctx context.Context, req *backend.CheckHealthRequest) co
 		return ctx
 	}
 
-	// TODO: Add support for ID and access tokens
+	// TODO: CheckHealthRequest currently doesn't support ID and access tokens, only user information
+	// See issue https://github.com/grafana/grafana-plugin-sdk-go/issues/579
 	currentUser := CurrentUserContext{
 		User:        req.PluginContext.User,
 		IdToken:     "",
@@ -57,7 +58,6 @@ func FromHealthCheckReq(ctx context.Context, req *backend.CheckHealthRequest) co
 }
 
 func getQueryReqHeader(req *backend.QueryDataRequest, headerName string) string {
-	// TODO: Check what subset of Unicode is case-insensitive for request headers, possibly narrow to ASCII
 	headerNameCI := strings.ToLower(headerName)
 
 	for name, value := range req.Headers {
@@ -70,7 +70,6 @@ func getQueryReqHeader(req *backend.QueryDataRequest, headerName string) string 
 }
 
 func getResourceReqHeader(req *backend.CallResourceRequest, headerName string) string {
-	// TODO: Check what subset of Unicode is case-insensitive for request headers, possibly narrow to ASCII
 	headerNameCI := strings.ToLower(headerName)
 
 	for name, values := range req.Headers {
--- a/pkg/azuredx/adxauth/adxusercontext/from_req.go
+++ b/pkg/azuredx/adxauth/adxusercontext/from_req.go
@@ -7,7 +7,7 @@ import (
 	"github.com/grafana/grafana-plugin-sdk-go/backend"
 )
 
-func FromQueryReq(ctx context.Context, req *backend.QueryDataRequest) context.Context {
+func WithUserFromQueryReq(ctx context.Context, req *backend.QueryDataRequest) context.Context {
 	if req == nil {
 		return ctx
 	}
@@ -24,7 +24,7 @@ func FromQueryReq(ctx context.Context, req *backend.QueryDataRequest) context.Co
 	return WithCurrentUser(ctx, currentUser)
 }
 
-func FromResourceReq(ctx context.Context, req *backend.CallResourceRequest) context.Context {
+func WithUserFromResourceReq(ctx context.Context, req *backend.CallResourceRequest) context.Context {
 	if req == nil {
 		return ctx
 	}
@@ -41,7 +41,7 @@ func FromResourceReq(ctx context.Context, req *backend.CallResourceRequest) cont
 	return WithCurrentUser(ctx, currentUser)
 }
 
-func FromHealthCheckReq(ctx context.Context, req *backend.CheckHealthRequest) context.Context {
+func WithUserFromHealthCheckReq(ctx context.Context, req *backend.CheckHealthRequest) context.Context {
 	if req == nil {
 		return ctx
 	}
--- a/pkg/azuredx/adxauth/adxusercontext/from_req.go
+++ b/pkg/azuredx/adxauth/adxusercontext/from_req.go
@@ -1,0 +1,98 @@
+package adxusercontext
+
+import (
+	"context"
+	"strings"
+
+	"github.com/grafana/grafana-plugin-sdk-go/backend"
+)
+
+func FromQueryReq(ctx context.Context, req *backend.QueryDataRequest) context.Context {
+	if req == nil {
+		return ctx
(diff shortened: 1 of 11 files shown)


pkg/azuredx/datasource.go
```diff
@@ -89,29 +87,27 @@ func NewDatasource(instanceSettings backend.DataSourceInstanceSettings) (instanc
 
 // QueryData is the primary method called by grafana-server
 func (adx *AzureDataExplorer) QueryData(ctx context.Context, req *backend.QueryDataRequest) (*backend.QueryDataResponse, error) {
+	ctx = adxusercontext.FromQueryReq(ctx, req)
```
andresmgot, line 90: Can you make it explicit that you are adding the tokens here? I would rename this to:

```
	ctx = adxusercontext.WithRequestToken(ctx, req)
```

So it's clear what's doing (and easier to follow).

pkg/azuredx/adxauth/adxusercontext/from_req.go
```diff
@@ -0,0 +1,98 @@
...
+
+	return WithCurrentUser(ctx, currentUser)
+}
+
+func FromHealthCheckReq(ctx context.Context, req *backend.CheckHealthRequest) context.Context {
+	if req == nil {
+		return ctx
+	}
+
+	// TODO: Add support for ID and access tokens
```
andresmgot, line 49: can you add a link here in the comment to the issue that is blocking this TODO?

The same applies to the other TODO comments that you have added. A link to an issue helps to know when the TODO can be resolved in the future.

pkg/azuredx/adxauth/adxusercontext/from_req.go
```diff
@@ -0,0 +1,98 @@
...
+		User:        req.PluginContext.User,
+		IdToken:     "",
+		AccessToken: "",
+	}
+
+	return WithCurrentUser(ctx, currentUser)
+}
+
+func getQueryReqHeader(req *backend.QueryDataRequest, headerName string) string {
+	// TODO: Check what subset of Unicode is case-insensitive for request headers, possibly narrow to ASCII
```
andresmgot, line 60: I would either remove this TODO or resolve it (doesn't sound like it's blocked by anything).

andresmgot (commented):
I am okay with the main change, my main concern is that we have both the `context` and the `backend.Requests` in the different methods and we are choosing to add the token info to the `context` rather than the `request` (which is what I would thought include the auth info at first glance) but it's probably not a big issue if we are explicit in the code about the decission.

Also, there is also the potential problem that the tokens may be pass down to functions that should not have access to them (but that have access to the context) but at the moment I don't see any example. What do you think?

pkg/azuredx/datasource.go
```diff
@@ -89,29 +87,27 @@ func NewDatasource(instanceSettings backend.DataSourceInstanceSettings) (instanc
...
 
-	authorization, err := adx.serviceCredentials.QueryDataAuthorization(ctx, req)
-	if err != nil {
-		return nil, err
-	}
+	res := backend.NewQueryDataResponse()
 
 	for _, q := range req.Queries {
-		res.Responses[q.RefID] = adx.handleQuery(q, req.PluginContext.User, authorization)
+		res.Responses[q.RefID] = adx.handleQuery(adxusercontext.FromQueryReq(ctx, req), q, req.PluginContext.User)
```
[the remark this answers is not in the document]
kostrse: Yes, this was overlooked. I will fix it.

pkg/azuredx/adxauth/adxusercontext/from_req.go
kostrse: Yes, will do

kostrse:
I created separate issue to remove remaining headers stuff out of `modelQuery`, is it won't need `backend.User` anymore:
* https://github.com/grafana/azure-data-explorer-datasource/issues/526

pkg/azuredx/adxauth/adxusercontext/from_req.go
kostrse: Yes, since these funcs are private and used only for specific headers `X-ID-Token`, `Authorization`, there's no need to worry about case sensitivity and Unicode.

pkg/azuredx/adxauth/adxusercontext/from_req.go
```diff
@@ -0,0 +1,98 @@
...
+	headerNameCI := strings.ToLower(headerName)
+
+	for name, value := range req.Headers {
+		if strings.ToLower(name) == headerNameCI {
+			return value
+		}
+	}
```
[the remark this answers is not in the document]
kostrse: Headers are case insensitive, I don't want to assume the case to not get regression in future.
So, I use case insensitive comparison.

pkg/azuredx/datasource.go
andresmgot: any comment about this name change?

kostrse: I don't like current naming too but I had to come up with something to have three distinct funcs: for query request, call resource request and health check request. If Go would support overloads, it would be much easier...

What about this?
```
ctx := adxusercontext.WithUserFromQueryReq(ctx, ...)

ctx := adxusercontext.WithUserFromResourceReq(ctx, ...)

ctx := adxusercontext.WithUserFromHealthCheckReq(ctx, ...)
```

`adx user context with user from query request` etc.

kostrse:
> There is one comment left and some spell / lint errors, let me know if you need any help with that
Tests fixed.

pkg/azuredx/datasource.go
andresmgot: yeah, that sounds good too (as far as it's clear that it's adding user / auth info to the context)

kostrse: Renamed.

andresmgot (approved):
LGTM, thanks for the changes!

PR #529, opened 2022-12-22
Common authentication settings UI

### LINKED ISSUE #361: Update frontend to common Azure Authentication UI and AzureCredentials format
Azure datasources have common design and implementation of configuration UI, common schema of AzureCredentials in JsonData and common implementation of parsers. This facilitates code reuse, easier changes across the board and consistency in user experience.

image

One of the features which are supported by other Azure datasources but not by Azure Data Explorer is Managed Identity (MSI) authentication.

As a precondition for implementation of MSI authentication, the Azure authentication section of configuration UI should be replaced by same implementation from other Azure datasources and credentials schema should be updated to same schema used by other datasources.

Here is the code which could be borrowed from Prometheus datasource.

Directory: [/public/app/plugins/datasource/prometheus/configuration](https://github.com/grafana/grafana/tree/main/public/app/plugins/datasource/prometheus/configuration)

* [AzureCredentialsForm.tsx](https://github.com/grafana/grafana/blob/main/public/app/plugins/datasource/prometheus/configuration/AzureCredentialsForm.tsx) - React component for Azure authentication configuration section
* [AzureAuthSettings.tsx#L20-L26](https://github.com/grafana/grafana/blob/b43e9b50b497f315f31fb35a006d820871d88dda/public/app/plugins/datasource/prometheus/configuration/AzureAuthSettings.tsx#L20-L26) - example of usage of AzureCredentialsForm
* [AzureCredentials.ts](https://github.com/grafana/grafana/blob/main/public/app/plugins/datasource/prometheus/configuration/AzureCredentials.ts) - domain objects, definitions of AzureCredentials
* [AzureCredentialsConfig.ts](https://github.com/grafana/grafana/blob/main/public/app/plugins/datasource/prometheus/configuration/AzureCredentialsConfig.ts) - reading/writing of AzureCredentials from/to JsonData 

Since ADX has existing credentials format, extra logic should be implemented which would read credentials in old format if credentials in new format were not found, so they would be rewritten in new format when saved.

This requires backend implementation first:
* #360

kostrse:
Azure authentication settings UI same as in Azure Monitor and Prometheus datasources.
Supports migration from legacy credentials format.

ADX_Creds

Fixes #361

### Migration of client secret

Since secret cannot be migrated, the existing legacy client secret field `clientSecret` is being preserved and will continued to be used until user updates the value. Once new value set, it will be saved to the new field `azureClientSecret`.

Users should not notice the fact of migration.

### Depends on
- https://github.com/grafana/grafana-azure-sdk-go/pull/23

### For future

The UI code can be moved into a shared JS library and shared between Azure Monitor, Prometheus and Azure Data Explorer.


src/components/ConfigEditor/AzureCredentialsForm.tsx
```diff
@@ -0,0 +1,294 @@
...
+        value: 'msi',
+        label: 'Managed Identity',
+      });
+    }
+
+    if (props.oboEnabled) {
+      opts.push({
+        value: 'clientsecret-obo',
+        label: 'App Registration (On-Behalf-Of)',
+      });
```
andresmgot, line 60: I prefer the current UI of having the "on-behalf-of" as a toggle for two reasons:

 - It's easier to notice that it's a variant of the "App Registration" auth mode.
 - It's easier to add the "Alpha" warning (which is lost here).

src/components/ConfigEditor/AzureCredentialsForm.tsx
```diff
@@ -0,0 +1,294 @@
...
+      };
+      onCredentialsChange(updated);
+    }
+  };
+
+  return (
+    <div className="gf-form-group">
+      {(authTypeOptions.length > 1) && (
+        <div className="gf-form-inline">
+          <div className="gf-form">
```
andresmgot, line 168: if we are going to rewrite this, please use the new componetns from `@grafana/ui`. Check the Azure Monitor config editor to see how it's done there (with `InlineField` components).

andresmgot (commented):
Thanks for the PR Sergey! I have several inline comments but in general, I think there are two things that we can do here:

 - Avoid the breaking change for the client secret. This means that we should support a partially migrated config, which is not going to be elegant from the code perspective but it would be nice for users.
 - Remove the default subscription code from this PR. I know this is to refactor this config editor with the Azure Monitor one but we can move that to a different component, so ADX doesn't depend on it.

src/components/ConfigEditor/AzureCredentialsConfig.ts
```diff
@@ -0,0 +1,197 @@
...
+  const jsonData = options.jsonData;
+
+  if (typeof jsonData.tenantId === 'string' || typeof jsonData.clientId === 'string') {
+    try {
+      const azureCloud = resolveLegacyCloudName(jsonData.cloudName);
+
+      return {
+        authType: jsonData.onBehalfOf ? 'clientsecret-obo' : 'clientsecret',
+        azureCloud: azureCloud,
+        tenantId: typeof jsonData.tenantId === 'string' ? jsonData.tenantId : '',
```
aangelisc, line 78: This type and `clientId` below shouldn't need to be re-validated.

src/components/ConfigEditor/AzureCredentialsForm.tsx
```diff
@@ -0,0 +1,294 @@
...
+  const hasRequiredFields = isCredentialsComplete(credentials);
+
+  const [subscriptions, setSubscriptions] = useState<Array<SelectableValue<string>>>([]);
+  const [loadSubscriptionsClicked, onLoadSubscriptions] = useReducer((val) => val + 1, 0);
+  useEffect(() => {
+    if (!getSubscriptions || !hasRequiredFields) {
+      updateSubscriptions([]);
+      return;
+    }
+    let canceled = false;
```
aangelisc, line 28: nit: `cancelled`

aangelisc (changes requested):
Nice work on this PR @kostrse. I echo @andresmgot comments but I actually think if we're making this change to better align the current Azure data sources then I would move this work to the [Azure React repo](https://github.com/grafana/grafana-azure-sdk-react). This will allow us to better maintain all common components required for Azure credentials. 

Wdyt? cc: @andresmgot

andresmgot:
> Nice work on this PR @kostrse. I echo @andresmgot comments but I actually think if we're making this change to better align the current Azure data sources then I would move this work to the [Azure React repo](https://github.com/grafana/grafana-azure-sdk-react). This will allow us to better maintain all common components required for Azure credentials.
> 
> Wdyt? cc: @andresmgot

yes, I believe this is an intermediate step before moving the code to the SDK

kostrse:
> > Nice work on this PR @kostrse. I echo @andresmgot comments but I actually think if we're making this change to better align the current Azure data sources then I would move this work to the [Azure React repo](https://github.com/grafana/grafana-azure-sdk-react). This will allow us to better maintain all common components required for Azure credentials.
> > Wdyt? cc: @andresmgot
> 
> yes, I believe this is an intermediate step before moving the code to the SDK

The end goal so to move everything to the shared library.

I would extract Ui to the shared library after Azure Monitor also migrated to the common 'azureCredentials' credentials format. I would look at Azure Monitor after Azure Data Explorer is done.

src/components/ConfigEditor/AzureCredentials.ts
```diff
@@ -0,0 +1,58 @@
...
+  { value: AzureCloud.USGovernment, label: 'Azure US Government' },
+] as SelectableValue[];
+
+export type AzureAuthType = 'msi' | 'clientsecret' | 'clientsecret-obo';
+
+export type ConcealedSecret = symbol;
+
+interface AzureCredentialsBase {
+  authType: AzureAuthType;
+  defaultSubscriptionId?: string;
```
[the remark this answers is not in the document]
kostrse: Agree, subscriptions in the credentials control only used in Azure Monitor ([here](https://github.com/grafana/grafana/blob/3c6e0e8ef85048af952367751e478c08342e17b4/public/app/plugins/datasource/grafana-azure-monitor-datasource/components/MonitorConfig.tsx#L40)).

I will create a work item to refactor it in Azure Monitor. We need all datasources to have same behavior so we could extract it into shared library.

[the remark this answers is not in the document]
kostrse: https://github.com/grafana/grafana/issues/60780

src/components/ConfigEditor/AzureCredentialsConfig.ts
```diff
@@ -0,0 +1,197 @@
...
+
+export function getDefaultCredentials(): AzureCredentials {
+  if (config.azure.managedIdentityEnabled) {
+    return { authType: 'msi' };
+  } else {
+    return { authType: 'clientsecret', azureCloud: getDefaultAzureCloud() };
+  }
+}
+
+export function getCredentials(options: DataSourceSettings<any, any>): AzureCredentials {
```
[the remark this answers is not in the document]
kostrse: I think that typing of `DataSourceJsonData` and `S` of `DataSourceSettings` in `@grafana/data` is inaccurate.

They are property bags and should allow key/value access but they are not typed as such. Sp I have to clear the type.

src/components/ConfigEditor/AzureCredentialsForm.tsx
kostrse: Done.

src/components/ConfigEditor/AzureCredentials.ts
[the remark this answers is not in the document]
kostrse: Removed `defaultSubscriptionId`, it will need to be addressed in Azure Monitor too.

src/components/ConfigEditor/AzureCredentials.ts
```diff
@@ -0,0 +1,58 @@
+import { SelectableValue } from '@grafana/data';
+
+export enum AzureCloud {
+  Public = 'AzureCloud',
+  China = 'AzureChinaCloud',
+  USGovernment = 'AzureUSGovernment',
+  None = '',
```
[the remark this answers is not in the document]
kostrse: Removed, not used.

src/components/ConfigEditor/AzureCredentialsConfig.ts
```diff
@@ -0,0 +1,197 @@
...
+      const azureCloud = resolveLegacyCloudName(jsonData.cloudName);
+
+      return {
+        authType: jsonData.onBehalfOf ? 'clientsecret-obo' : 'clientsecret',
+        azureCloud: azureCloud,
+        tenantId: typeof jsonData.tenantId === 'string' ? jsonData.tenantId : '',
+        clientId: typeof jsonData.clientId === 'string' ? jsonData.clientId : '',
+        // When converting existing credentials from the legacy format,
+        // client secret is going to be lost and will need to be reentered
+        clientSecret: undefined,
```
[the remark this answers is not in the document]
kostrse: Updated the code. On frontend I now use two instances of symbol to pass information whether the new or legacy field should be preserved.

This is the change for backend:
- https://github.com/grafana/grafana-azure-sdk-go/pull/23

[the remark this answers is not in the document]
kostrse: Added tests covering client secret handling.

src/components/ConfigEditor/AzureCredentialsConfig.ts
```diff
@@ -0,0 +1,197 @@
...
+    case 'chinaazuremonitor':
+      return AzureCloud.China;
+    case 'govazuremonitor':
+      return AzureCloud.USGovernment;
+    default:
+      throw new Error(`Azure cloud '${cloudName}' not supported by Azure Data Explorer datasource.`);
+  }
+}
+
+export function updateCredentials(
```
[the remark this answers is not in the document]
kostrse: **updateCredentials** means **_save credentials_** here, this is the setter method of pair **getCredentials**/**updateCredentials**.
It doesn't mean update to a new format. This is the semantic which is already used in all other datasources.

The actual storage format is implementation detail and not exposed outside. Caller always gets the same `AzureCredentials` object, regardless whether it was stored in legacy or new format.

src/components/ConfigEditor/AzureCredentialsConfig.ts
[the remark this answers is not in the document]
kostrse: Created an issue #530 to address it separately. This is a big issue which needs additional consideration.

src/components/ConfigEditor/AzureCredentialsForm.tsx
kostrse: `canceled` and `cancelled` are equally correct. I usually see `cancel/canceled` but `cancelling/cancellation`.

Example: https://learn.microsoft.com/en-us/dotnet/api/system.threading.cancellationtoken.canbecanceled?view=net-7.0

kostrse: I will keep as is since it's already used in other datasources.

kostrse: It was removed anyways.

src/components/ConfigEditor/AzureCredentialsConfig.ts
kostrse: The type needs to be re-evaluated because the initial condition is `OR`, it doesn't guarantee that both fields are string:

```go
typeof jsonData.tenantId === 'string' || typeof jsonData.clientId === 'string'
```

I use `OR` because even if the existing credentials are incomplete and some input is missing, I still attempt to migrate whichever is present, as opposed to having everything blank because some fields were missing.

kostrse:
I would want OBO to be considered a distinct authentication type, not a subset of App Registration.
One is service identity, another is user identity, substantial difference.



Here is the warning banner for OBO:

OBO in Beta

Please let me know what needs to be changed in the wording of the message.
Also, do we need a link to documentation? What would the the best target for it then?

src/components/ConfigEditor/AzureCredentials.ts
```diff
@@ -0,0 +1,56 @@
...
+  USGovernment = 'AzureUSGovernment',
+}
+
+export const KnownAzureClouds = [
+  { value: AzureCloud.Public, label: 'Azure' },
+  { value: AzureCloud.China, label: 'Azure China' },
+  { value: AzureCloud.USGovernment, label: 'Azure US Government' },
+] as SelectableValue[];
+
+export type AzureAuthType = 'msi' | 'clientsecret' | 'clientsecret-obo';
```
andresmgot, line 15: nit: This can be an `enum` too

src/components/ConfigEditor/AzureCredentialsConfig.ts
```diff
@@ -0,0 +1,207 @@
+import { DataSourceSettings } from '@grafana/data';
+import { config } from '@grafana/runtime';
+
+import { AzureCloud, AzureCredentials, ConcealedSecret } from './AzureCredentials';
+
+const concealed: ConcealedSecret = Symbol('Concealed client secret');
+const concealedLegacy: ConcealedSecret = Symbol('Concealed legacy client secret');
```
andresmgot, line 7: is there any benefit of doing this distinction? AFAIK the logic will be the same.

src/components/ConfigEditor/AzureCredentialsConfig.ts
```diff
@@ -0,0 +1,207 @@
...
+          ...options.secureJsonData,
+          azureClientSecret:
+            typeof credentials.clientSecret === 'string' && credentials.clientSecret.length > 0
+              ? credentials.clientSecret
+              : undefined,
+        },
+        secureJsonFields: {
+          ...options.secureJsonFields,
+          azureClientSecret: credentials.clientSecret === concealed,
+          clientSecret: credentials.clientSecret === concealedLegacy,
```
andresmgot, line 177: what happens if you don't update the legacy way of setting this?

src/components/ConfigEditor/AzureCredentialsForm.tsx
```diff
@@ -0,0 +1,218 @@
...
+  }, [props.managedIdentityEnabled, props.oboEnabled]);
+
+  const onAuthTypeChange = (selected: SelectableValue<AzureAuthType>) => {
+    if (onCredentialsChange) {
+      const updated: AzureCredentials = {
+        ...credentials,
+        authType: selected.value || 'msi',
+      };
+      onCredentialsChange(updated);
+    }
```
andresmgot, line 51: nit:
```suggestion
    if (onCredentialsChange && selected.value) {
      const updated: AzureCredentials = {
        ...credentials,
        authType: selected.value,
      };
      onCredentialsChange(updated);
    }
```

src/components/ConfigEditor/AzureCredentialsForm.tsx
```diff
@@ -0,0 +1,218 @@
...
+            className="width-15"
+            value={authTypeOptions.find((opt) => opt.value === credentials.authType)}
+            options={authTypeOptions}
+            onChange={onAuthTypeChange}
+          />
+        </InlineField>
+      )}
+      {credentials.authType === 'clientsecret-obo' && (
+        <>
+          {/* eslint-disable prettier/prettier */}
```
andresmgot, line 139: why disabling prettier?

src/tracking.ts
```diff
@@ -89,9 +86,6 @@ export const analyzeQueries = (queries: KustoQuery[], datasourceSrv: DataSourceS
       datasources[JSON.stringify(query.datasource)] = dsSettings;
     }
     if (dsSettings) {
-      if (dsSettings.jsonData?.onBehalfOf) {
-        counters.on_behalf_of_queries++;
-      }
```
andresmgot, line 94: this information is still useful, you can check if the `authType` is  `clientsecret-obo` instead.

andresmgot (commented):
Thanks for the changes, I have tested the changes and it's working for me, I just have some minor comments.

andresmgot:
> Please let me know what needs to be changed in the wording of the message.
> Also, do we need a link to documentation? What would the the best target for it then?

I think this is fine then, I would just change the text of the link.

src/components/ConfigEditor/AzureCredentialsConfig.ts
kostrse: `clientSecret` of `AzureClientSecretCredentials` has type `string | symbol`.

the `string` is when user changed the secret and actual secret stored in the string. the `symbol` is when we know that secret is stored on backend but frontend doesn't have its actual value.

For outside code its only either `string` or `symbol`. The `concealed` and `concealedLegacy` variables are not exported outside the module and used internally to handle the legacy scenario.

Newly created credentials store secret in the `azureClientSecret` field of `SecureJsonData`, legacy credentials store in `clientSecret` field. When `getCredentials` function reads credentials from JsonData/SecureJsonData, it returns `AzureCredentials` object and information about which field was used is being lost.

When `updateCredentials` method writes the `AzureCredentials` object back to JsonData/SecureJsonData and secret wasn't changed, then it should preserve the whatever the original field was used. So difference between `concealed` and `concealedLegacy` is information that tells the `updateCredentials` method whether the  `azureClientSecret` or `clientSecret` field was originally used.

This information known only to the module, for external code it's just `symbol` with no any difference.

src/components/ConfigEditor/AzureCredentialsConfig.ts
kostrse: Assume that existing credentials stored in legacy format, existing secret stored in the `clientSecret` field on backend.

User goes to the datasource configuration UI, changes only `clientId` and clicks Save. Secret wasn't changed by the user and expected to be kept intact.

Credentials can only be saved in the new format, so the updated credentials will be stored in the new format, `azureCredentials` object. The new format normally keeps secret in another field - `azureClientSecret`. Since user didn't update the secret, the existing secret should not be lost. So this code has extra handling of this legacy secret scenario. If secret stored in `azureClientSecret` then keep it in `azureClientSecret`, if secret already stored in `clientSecret` then continue to keep it there.

These tests are covering this scenario:
https://github.com/kostrse/azure-data-explorer-datasource/blob/f104f5ccd24da0cee568c92bcb56ad37f47a67ac/src/components/ConfigEditor/AzureCredentialsConfig.test.ts#L100

src/tracking.ts
kostrse: Tracking was restored.

src/components/ConfigEditor/AzureCredentialsForm.tsx
kostrse: Prettier is enforcing complete nonsense here

kostrse: It doesn't like the plain text here and tries to format it in a weird way

src/components/ConfigEditor/AzureCredentials.ts
kostrse: it's functionally the same but already defined this way in other datasources.

src/components/ConfigEditor/AzureCredentialsConfig.ts
```diff
@@ -0,0 +1,197 @@
...
+        // then we should fall back to an empty default credentials
+        return getDefaultCredentials();
+      }
+    case 'clientsecret':
+    case 'clientsecret-obo':
+      if (credentials.authType === 'clientsecret-obo' && !getOboEnabled()) {
+        // If authentication type is OBO but OBO were disabled in Grafana config,
+        // then we should fall back to an empty default credentials
+        return getDefaultCredentials();
+      }
```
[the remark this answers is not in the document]
kostrse: My IDE (ESLint?) doesn't like it. And I agree, it's too dangerous, it can be accidentally changed in future, someone may rearrange the order of cases without noticing that this is the fall thru, and everything will get broken. Let's keep it more straightforward. even if it's checking same thing two times.

andresmgot (approved):
LGTM, @aangelisc do you want to do a final check too?

BTW, this will break the e2e in `main` because the UI changes but I can send a quick PR to fix it.

andresmgot:
> This all looks good to me as well. R.E. E2E tests, @kostrse would you be able to re-add the `data-testid` props that were in the original `ConfigEditor`? This should be a straightforward change that will ensure the E2E tests do not break. If not we can add another PR for this later blush

There is one extra change needed, apart from the test-id. I have prepared a quick PR we can merge after this.

PR #531, opened 2023-01-06
Update secret

aangelisc:
Switch to using `GITHUB_TOKEN`


asimpson (approved):
Approving so you can merge when you're ready since we're on this call together.

PR #536, opened 2023-01-10
Prepare v4.2.0 release

### LINKED ISSUE #391: Revamp query builder

andresmgot:
Closes #391

I am still leaving the legacy editor behind a feature flag to be able to rollback to the previous editor in case there is some unexpected bug.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Change Log
 
+## [4.2.0]
+
+This release enables a new query builder:
+
+- New feature: It's now possible to filter columns to query, improving the performance of queries.
+- The query preview includes syntax highlighting for Kusto.
+- The rest of components has been refactored to match the latest Grafana UI.
+
 ## [4.1.10]
 
 - Fix: Invalid code editor loaded for Grafana versions that don't follow semantic versioning by @aangelisc in https://github.com/grafana/azure-data-explorer-datasource/pull/506
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,15 @@
 
 This release revamps the plugin query builder:
 
-- New feature: It's now possible to filter columns to query, improving the performance of queries.
+- New feature: It's now possible to filter columns within a query, improving the performance of queries.
 - The query preview includes syntax highlighting for Kusto.
-- The rest of components has been refactored to match the latest Grafana UI.
+- All other components have been refactored to match the latest Grafana UI.
+
+Apart from that, this release also includes:
+
+- Refactor: Authentication and configuration has been refactored to match other Azure plugins.
+- Fix: Health check for data sources configured with On-Behalf-Of authentication.
+- Fix: Alert queries that returns no data.
 
 ## [4.1.10]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [4.2.0]
 
-This release enables a new query builder:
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Change Log
 
+## [4.2.0]
+
+This release revamps the plugin query builder:
+
+- New feature: It's now possible to filter columns to query, improving the performance of queries.
```
aangelisc, line 7:
```suggestion
- New feature: It's now possible to filter columns within a query, improving the performance of queries.
```

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
...
+- The query preview includes syntax highlighting for Kusto.
+- The rest of components has been refactored to match the latest Grafana UI.
```
aangelisc, line 9: nit:
```suggestion
- All other components have been refactored to match the latest Grafana UI.
```

src/components/QueryEditor/QueryHeader.tsx
```diff
@@ -110,7 +111,7 @@ export const QueryHeader = (props: QueryEditorHeaderProps) => {
         }}
       />
       <FlexItem grow={1} />
-      <Button variant="primary" icon="play" size="sm" onClick={onRunQuery}>
+      <Button variant="primary" icon="play" size="sm" onClick={onRunQuery} aria-label="Run query">
```
aangelisc, line 114: Are `aria-label`'s needed here given that these components have text labels already?

aangelisc (approved):
LGTM, just some questions/nits in the comments 😊

src/components/QueryEditor/QueryHeader.tsx
andresmgot: yes, at least for the e2e tests to find them as it was in the legacy editor

andresmgot:
thanks for the review @aangelisc! I added a couple more items, also covering #529 so let's wait a bit to include that PR in the release

src/components/QueryEditor/QueryHeader.tsx
aangelisc: I thought this might be why 😊 It may be prudent to move these new labels to `data-testid` for consistency, wdyt?

andresmgot: Previously it was using aria-labels but we can change it. It seems it doesn't work with the `InlineSelect` (it does not forward the `data-testid`) but it does for the `Button`.

aangelisc: I've opened an issue [here](https://github.com/grafana/grafana-experimental/issues/54) for the `InlineSelect` component 😊

asimpson (approved):

PR #540, opened 2023-01-13
Adapt e2e tests

andresmgot:
Needed after #529

### DIFF
--- a/cypress/integration/smoke.spec.ts
+++ b/cypress/integration/smoke.spec.ts
@@ -26,7 +26,7 @@ function addCommonProvisioningADXDatasource(ADXProvisions: ADXProvision[]) {
   e2e.flows.addDataSource({
     type: 'Azure Data Explorer Datasource',
     form: () => {
-      e2eSelectors.configEditor.azureCloud.input().find('#azure-cloud-type').type('Azure');
+      e2eSelectors.configEditor.azureCloud.input().find('input').type('Azure');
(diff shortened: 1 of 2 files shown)


src/components/ConfigEditor/AzureCredentialsForm.tsx
```diff
@@ -136,18 +137,29 @@ export const AzureCredentialsForm: FunctionComponent<Props> = (props: Props) =>
       )}
       {credentials.authType === 'clientsecret-obo' && (
         <>
-          {/* eslint-disable prettier/prettier */}
```
andresmgot, line 139: my editor was ignoring this so the file was re-formatted anyway.

cypress/integration/smoke.spec.ts
```diff
@@ -26,7 +26,7 @@ function addCommonProvisioningADXDatasource(ADXProvisions: ADXProvision[]) {
   e2e.flows.addDataSource({
     type: 'Azure Data Explorer Datasource',
     form: () => {
-      e2eSelectors.configEditor.azureCloud.input().type('Azure');
+      e2eSelectors.configEditor.azureCloud.input().find('#azure-cloud-type').type('Azure');
```
aangelisc, line 29: nit: I think `.find('input')` here is more stable rather than referencing the `htmlFor` field

aangelisc (approved):
👍

Issue #486, opened 2022-10-10
P3: When adding a filter/aggregate/groupby: keyboard focus should be on first select


yaelleC:
When using the keyboard the create an ADX query, and hitting the `+` button to create a filter/agregate/group by item, the focus is lost. 
It would be nice to have focus on the first element of the newly created form instead.

We have set up this issue as lowest a11y prio as it requires quite a lot of effort 

https://user-images.githubusercontent.com/42030685/194829786-988eeb6f-472d-419f-b9b5-48881fee9c41.mov

To Do

- [x] Get UX input to confirm the wanted behaviour for the form
- [ ] Implement the changes

andresmgot:
does this happens with the new editor? (this editor will go away)

asimpson:
@andresmgot Yep, it also happens on the new Editor. 

@diegoadams or @catherineymgui any thoughts from a UX perspective here?

catherineymgui:
Yeah, I agree that there should be focus on the first element once a new filter is created.

PR #543, opened 2023-01-19
Puts focus on field when a new field is added.

asimpson:
Peek 2023-01-19 10-08

**To test**
1. Load up ADX
2. Select the correct database
3. Navigate from that dropdown to the Builder via `TAB` on your keyboard.
4. You should be able to add filter groups and and other clauses via `TAB` and `Enter` or `Space`. The `focus` should move to the logically correct place after taking an action.

**Other changes**
- Fixes various eslint issues. We might want to revisit how [`@grafana/experimental` defines `Children`](https://github.com/grafana/grafana-experimental/blob/2ce409b4c94f68f4789b30430d01d6f77c56ccc0/src/QueryEditor/InputGroup.tsx#L5) to avoid the `Fragment` usage here?
- I also could see a desire to separate "ui" properties from data properties in our types, e.g. `focus` being added to `QueryEditorReduceExpressionArray`.
- Updates grafana/experimental version

Ref #486


asimpson:
fixing tests...

aangelisc (approved):
LGTM 👍

alyssabull (approved):
LGTM!

PR #544, opened 2023-01-19
Fix label width on ConfigEditor

asimpson:
**Before**
image

**After**
image


aangelisc (approved):
I love a simple fix 😊

PR #569, opened 2023-04-22
User identity authentication in configuration UI

### LINKED ISSUE #381: Add user identity authentication option to configuration UI
Once Grafana Azure SDK supports user identity authentication, add support for it in frontend:
* Extend AzureCredentials struct and JsonData readers/writers
* Add option to Configuration UI

kostrse:
This PR adds "Current User" as an authentication option for the datasource.

Current User authentication

The option is available to the user only if user identity authentication was enabled in Grafana config (see https://github.com/grafana/grafana/pull/50277). It can be enabled as default via feature flag `adxUserIdentityPreferred `.

Fixes #381

### DIFF
--- a/doc/current-user-auth.md
+++ b/doc/current-user-auth.md
@@ -1,0 +1,57 @@
+# Current User Authorization
+
+**Note**: The feature is experimental, which means it may not work as expected and breaking changes may be introduced
+in the future.
+
+⚠️ Only compatible with Grafana 10.0 or above.
+
+## Prerequisites
+
+Current User Authorization only can work together with Azure AD Grafana authentication.
+
+## Configuration
+
+Assuming that Azure AD authentication has been already configured, for example:
+
+```ini
+[auth.azuread]
+enabled = true
+
+auth_url = https://login.microsoftonline.com/fd719c11-a91c-40fd-8379-1e6cd3c59568//oauth2/v2.0/authorize
+token_url = https://login.microsoftonline.com/fd719c11-a91c-40fd-8379-1e6cd3c59568/oauth2/v2.0/token
+client_id = f85aa887-490d-4fac-9306-9b99ad0aa31d
+client_secret = 87808761-ff7b-492e-bb0d-5de2437ffa55
+```
+
+Current User authentication can be enabled in the `[azure]` section of the Grafana config:
+
+```ini
+[azure]
+user_identity_enabled = true
+```
+
+Optionally, it's possible to provide another AAD app for token exchange (On-Behalf-Of token request):
+
+```ini
+[azure]
+user_identity_enabled = true
+
+# Use different AAD app (OBO)
+user_identity_client_id = 4fc34037-97bd-4e84-9db4-86238c78e32a
+user_identity_client_secret = 4479f5a6-444c-4271-8790-60eeb42225ae
+```
+
+Also, it's possible to customize the token endpoint:
+
+```ini
+[azure]
+user_identity_enabled = true
+
+user_identity_token_url = https://custom-token-endpoint/oauth2/v2.0/token
+user_identity_client_id = 4fc34037-97bd-4e84-9db4-86238c78e32a
+user_identity_client_secret = 4479f5a6-444c-4271-8790-60eeb42225ae
+```
+
+## See also
+
+- [User identity authentication in Azure Data Explorer and other Azure datasources](https://github.com/grafana/grafana/discussions/62994)
--- a/doc/current-user-auth.md
+++ b/doc/current-user-auth.md
@@ -1,6 +1,6 @@
 # Current User Authorization
 
-**Note**: The feature is experimental, which means it may not work as expected and breaking changes may be introduced
+**Note**: The feature is experimental, which means it may not work as expected, it may cause Grafana to behave in an unexpected way, and breaking changes may be introduced
 in the future.
 
 ⚠️ Only compatible with Grafana 10.0 or above.
--- a/src/components/ConfigEditor/AzureCredentialsForm.tsx
+++ b/src/components/ConfigEditor/AzureCredentialsForm.tsx
@@ -7,6 +7,7 @@ import { AzureAuthType, AzureCredentials } from './AzureCredentials';
 import { selectors } from 'test/selectors';
 
 export interface Props {
+  userIdentityEnabled: boolean;
   managedIdentityEnabled: boolean;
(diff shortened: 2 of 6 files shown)


aangelisc (changes requested):
Initial testing on this with grafana/grafana-azure-sdk-go#32 and grafana/grafana#50277 has been unsuccessful. With the necessary updates to appropriately configure user-based auth there are some issues:

- The default auth is set to `Current User`, however with the `clusterUrl` value set the schema cannot be reloaded until the auth type option has been reselected.
- Attempting to reload the schema leads to the error: `invalid Azure configuration: user identity authentication is not supported by this datasource`

The only configuration I've added is to set `user_auth_enabled` to `true`. All other parameters needed should be pulled from the existing `auth.azuread` configuration for OBO.

Additionally, setting the `authType` to `Current User` should present the user with an alert describing the feature as experimental and highlighting that certain features will be broken, particularly alerting. An `InlineBanner` component should suffice for the details here.

kostrse:
> * The default auth is set to `Current User`, however with the `clusterUrl` value set the schema cannot be reloaded until the auth type option has been reselected.

I will double check, if there's some UI-specific bug in the behavior of the form, I will see what can be done.

> * Attempting to reload the schema leads to the error: `invalid Azure configuration: user identity authentication is not supported by this datasource`

I think because you need this in `httpclient.go` (after [this](https://github.com/kostrse/azure-data-explorer-datasource/blob/dc613e1eb23d7ede6568f77a7207503e6c847dcb/pkg/azuredx/client/httpclient.go#L18) line):

```go
	// Allow user identity authentication for this datasource if it's configured in Grafana config
	authOpts.AllowUserIdentity()
```

Datasource needs to explicitly enable support for user identity auth. This is security precaution, e.g. if a datasource isn't designed to properly isolate cache etc., then user identity auth should not be possible.

Plus add 3rd argument 'false' to `token_provider_azure.go` to make Grafana core compile ([here](https://github.com/grafana/grafana/blob/69402f859b314dfdf37b6535316f7a86b6eac1e3/pkg/api/pluginproxy/token_provider_azure.go#L23)).

These changes are not included into existing PRs because they need to be introduced in correct order and will be added later.

> 
> The only configuration I've added is to set `user_auth_enabled` to `true`. All other parameters needed should be pulled from the existing `auth.azuread` configuration for OBO.

If you add the `authOpts.AllowUserIdentity()` above, only `user_auth_enabled` to `true` should be enough.

> Additionally, setting the `authType` to `Current User` should present the user with an alert describing the feature as experimental and highlighting that certain features will be broken, particularly alerting. An `InlineBanner` component should suffice for the details here.

I will add alert message.

I hope you should be able to make it work end to end.

Each PR is self-containing and responsible for specific parts, so we can get back to this UI PR later, after we finish the backend part.

The first PR which I want merge is this one:
https://github.com/grafana/grafana-azure-sdk-go/pull/32

This will allow me to release a new version of SDK

kostrse:
Added the experimental banner and [documentation page](https://github.com/grafana/azure-data-explorer-datasource/blob/fecec07da8d08beefa06fdbd70f112f78d9320fa/doc/current-user-auth.md).

JsonData:
```json
{
    "azureCredentials":
    {
        "authType": "currentuser"
    },
    "clusterUrl": "https://example.kusto.windows.net",
    "dataConsistency": "strongconsistency",
    "defaultEditorMode": "visual",
    "oauthPassThru": true
}
```

doc/current-user-auth.md
```diff
@@ -0,0 +1,57 @@
+# Current User Authorization
+
+**Note**: The feature is experimental, which means it may not work as expected and breaking changes may be introduced
```
aangelisc, line 3:
```suggestion
**Note**: The feature is experimental, which means it may not work as expected, it may cause Grafana to behave in an unexpected way, and breaking changes may be introduced
```

aangelisc: I would also add a note admonishing what we've already identified r.e. issues with backend features.

src/components/ConfigEditor/AzureCredentialsForm.tsx
```diff
@@ -135,6 +143,21 @@ export const AzureCredentialsForm: FunctionComponent<Props> = (props: Props) =>
           />
         </InlineField>
       )}
+      {credentials.authType === 'currentuser' && (
+        <>
+          <Alert title="Current user authentication is experimental" severity="warning">
+            For known limitations and issues, bug report, or feedback, please visit{' '}
```
aangelisc, line 149:
```suggestion
            Certain Grafana features (e.g. alerting) may not work as expected. For other known limitations and issues, bug reports, or feedback, please visit{' '}
```

aangelisc (changes requested):
I've added some additional comments on the wording. Ideally, we'd like to be as specific as possible with what may or may not work right now so that we don't confuse users and cause ourselves to receive unexpected issue reports. 

Also, the form still doesn't behave as expected because the `hasCredentials` function needs to be updated to take into account the fact that current user auth does not require any additional properties 😊

kostrse:
> Also, the form still doesn't behave as expected because the `hasCredentials` function needs to be updated to take into account the fact that current user auth does not require any additional properties

Logic of `hasCredentials` from [here](https://github.com/grafana/azure-data-explorer-datasource/blob/4031dabf6f7a0a03cde45ef672797554f9e46e64/src/components/ConfigEditor/AzureCredentialsConfig.ts#LL38C1-L43C2):
```ts
    typeof options.jsonData.azureCredentials === 'object' ||
    (typeof options.jsonData.tenantId === 'string' && typeof options.jsonData.clientId === 'string')
```

If `azureCredentials` object exists (of any type, including user auth), then credentials are set, otherwise check the legacy scenario when client fields were set directly in the root of `JsonData`, not in `azureCredentials` object.

This considered as having credentials:
```json
{
  "azureCredentials": {
    "..."
  }
}
```

aangelisc:
Yep I checked the logic beforehand. The problem is when a user creates a new datasource the current user auth method is the default method but it is not actually set in the `jsonData` value which causes the reload button to be disabled.

kostrse:
> Yep I checked the logic beforehand. The problem is when a user creates a new datasource the current user auth method is the default method but it is not actually set in the `jsonData` value which causes the reload button to be disabled.

This is a separate bug, not specific to user aurh.

aangelisc (approved):

PR #574, opened 2023-05-22
Convert number to bool

bossinc:
Fixes https://github.com/grafana/support-escalations/issues/5928
Fixes issue where
```
.show queries
| extend B = true
| limit 1
```
returns a cannot convert error because it is getting a number when it expects a bool. The converter can now accept a bool or number.

Also reformatted a json file.

### DIFF
--- a/pkg/azuredx/models/table.go
+++ b/pkg/azuredx/models/table.go
@@ -214,7 +214,7 @@ var boolConverter = data.FieldConverter{
 			b = i != "0"
 			return &b, nil
 		}
-		return nil, fmt.Errorf("unexpected type, expected bool or json.Number but got got type %T with a value of %v", v, v)
+                return nil, fmt.Errorf("unexpected type, expected bool or json.Number but got type %T with a value of %v", v, v)
 	},
 }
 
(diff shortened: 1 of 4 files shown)


pkg/azuredx/models/table.go
```diff
@@ -206,10 +206,15 @@ var boolConverter = data.FieldConverter{
...
+		if ok {
+			return &b, nil
+		}
+		i, ok := v.(json.Number)
+		if ok {
+			b = i != "0"
+			return &b, nil
 		}
-		return &b, nil
+		return nil, fmt.Errorf("unexpected type, expected bool or json.Number but got got type %T with a value of %v", v, v)
```
asimpson, line 217:
```suggestion 
                return nil, fmt.Errorf("unexpected type, expected bool or json.Number but got type %T with a value of %v", v, v)
```

asimpson (approved):
Looks good, just one typo.

aangelisc:
I'd hold off merging this PR for now. I want to confirm that this isn't a bug in how ADX interprets queries because the following query works fine:

```kql
PerfTest
| extend b = true
| limit 1
```

alyssabull (approved):

aangelisc (approved):
This is good to go 😊 There is a subtle difference in the management vs query APIs which is why this exists.

PR #576, opened 2023-05-23
Set default credentials when datasource created

### LINKED ISSUE #575: Default auth type is not present in json data
The default auth type is not present in JSON data and does not get updated until `onChange` is called by either updating the auth type or editing credential information.

This affects the roll-out of user-based auth which will become the default authentication type. The json data should be appropriately created with the default auth type.

kostrse:
This PR tries to fix a problem when a newly created datasource has no JSON data by default.

It utilizes `useEffect` to preset default credentials configuration if credentials haven't been configured.

Fixes #575

Notes: I introduced `updateOptions` function which is similar to `updateJsonData` but more generic, so it's usable with `updateCredentials`. It's needed to avoid possible code racing when `useEffect` body is being called only in upcoming render cycle. I would prefer other code to also use `updateOptions` instead of `updateJsonData`.

### DIFF
--- a/src/components/ConfigEditor/index.tsx
+++ b/src/components/ConfigEditor/index.tsx
@@ -1,14 +1,20 @@
-import React, { useCallback, useMemo } from 'react';
+import React, { useCallback, useEffect, useMemo } from 'react';
 import { DataSourcePluginOptionsEditorProps } from '@grafana/data';
 import { config } from '@grafana/runtime';
 import ConfigHelp from './ConfigHelp';
-import { AdxDataSourceOptions, AdxDataSourceSecureOptions } from 'types';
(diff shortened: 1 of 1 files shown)


aangelisc:
This looks good @kostrse, should we also update `getDefaultCredentials` to handle the user auth case?

aangelisc (approved):

asimpson (approved):

kostrse:
> This looks good @kostrse, should we also update `getDefaultCredentials` to handle the user auth case?

`getDefaultCredentials` is updated [here](https://github.com/kostrse/azure-data-explorer-datasource/blob/8ec4096866dd38d8965b800b20fab4f92c8bc278/src/components/ConfigEditor/AzureCredentialsConfig.ts#L47-L48) in #569 

This PR is not specific to user auth.

PR #577, opened 2023-05-24
Add OpenAI integration

asimpson:
This integration adds an additional field in the datasource Config where the user can save their OpenAI API token. Once that token is present the user can use an additional mode of the Editor where they can use natural language to generate KQL queries.

**Additional future enhancements could be:**

- [ ] Display cost estimate before generating query
- [ ] Give the user the option to include the schema in the prompt to get more personalized results

**Examples**
openai-adx
openai-adx-advanced

### DIFF
--- a/cspell.config.json
+++ b/cspell.config.json
@@ -160,6 +160,9 @@
     "retrievebytes",
     "pascaldekloe",
     "xorcare",
+    "openai",
+    "AIAPI",
+    "AIURL",
     ""
   ]
 }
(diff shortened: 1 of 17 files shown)


aangelisc (approved):
Cool as hell, straightforward, great stuff 💯

alyssabull (approved):
This is awesome! Great work 💪

bossinc (approved):

PR #581, opened 2023-06-08
Improved tracking

aangelisc:
This PR adds additional tracking for the following:

- OpenAI queries
- Current user auth queries
- MSI queries
- App registration queries

This should provide us with greater insight into what authentication type queries use. It will also provide us with a more complete picture of what type of queries are issued (with all query types now covered). 

An additional supporting type has been added and the `AdxDatasourceOptions` type has been updated to include missing properties.


asimpson (approved):
Ah this is solid man :clap:

PR #585, opened 2023-06-09
OpenAI: UX changes per feedback from UX team

asimpson:
The changes here are largely about aligning the buttons with their respective text areas. The other change is hiding the run button in the header when in the OpenAI editor. See screenshots  below for before and after:

**Before**

image

**After**

image


aangelisc (approved):
LGTM 😊

PR #593, opened 2023-06-29
Fix: Check if OpenAi token is missing

asimpson:
Checks if the user has added their OpenAI token to the datasource configuration before we make a request with no token.


bossinc (approved):
🥳

PR #597, opened 2023-07-14
Fix bug that preventing switching back to ADX datasource

alyssabull:
This PR fixes a bug that threw an error when switching back to an ADX datasource from a different datasource.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,6 @@
     "@grafana/experimental": "^1.1.0",
     "@grafana/runtime": "9.2.1",
     "@grafana/ui": "9.2.1",
-    "@testing-library/dom": "^9.3.1",
     "debounce-promise": "^3.1.2",
     "react": "17.0.2",
     "react-dom": "17.0.2",
@@ -96,6 +95,6 @@
     "uuid": "^9.0.0"
   },
   "resolutions": {
-    "**/@testing-library/dom": "8.0.0"
+    "**/@testing-library/dom": "7.31.2"
   }
 }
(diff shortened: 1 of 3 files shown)


aangelisc (changes requested):
Because of overlap between the `KustoQuery` type and `ExpressionQuery` this change will cause the page to crash if an expression query is present and the user tries to switch to an ADX datasource. This is because of the addition of `query.expression` in the condition for rendering of the `VisualQueryEditor`. 

I think maybe we can just remove that requirement from the condition to make sure this still works as expected.

package.json
```diff
@@ -87,6 +87,7 @@
     "@grafana/experimental": "^1.1.0",
     "@grafana/runtime": "9.2.1",
     "@grafana/ui": "9.2.1",
+    "@testing-library/dom": "^9.3.1",
```
asimpson, line 90: I'm not sure why this got added here? If you back out that change and re-run `yarn install --pure-lockfile` what happens?

asimpson (changes requested):

aangelisc (approved):

yarn.lock
```diff
@@ -2828,6 +2828,20 @@
     lz-string "^1.4.4"
     pretty-format "^26.6.2"
 
+"@testing-library/dom@^9.3.1":
```
asimpson, line 2831: we  don't need these changes now since the package.json changes went away.

asimpson (changes requested):

yarn.lock
alyssabull: Done - ty!

asimpson (approved):
Woot, nicely done, this one was tricky :trophy:

PR #606, opened 2023-08-14
Update Grafana Azure SDK to 1.8

kostrse:
Grafana Azure SDK has updated dependencies to Azure SDK.


pkg/azuredx/adxauth/aad_client.go
```diff
@@ -16,7 +16,7 @@ import (
 
 // Abstraction over confidential.Client from MSAL for Go
 type aadClient interface {
-	AcquireTokenOnBehalfOf(ctx context.Context, userAssertion string, scopes []string) (confidential.AuthResult, error)
+	AcquireTokenOnBehalfOf(ctx context.Context, userAssertion string, scopes []string, options ...confidential.AcquireOnBehalfOfOption) (confidential.AuthResult, error)
```
aangelisc, line 19: Is this the breaking change @kostrse?

kostrse: Not in Grafana Azure SDK.   Grafana Azure SDK references a newer version of Azure SDK as a dependency, which has some contracts changed. So when updating Grafana Azure SDK,   Azure SDK as intermediate dependency needs to be updated as well.

aangelisc (approved):

PR #609, opened 2023-08-22
Update plugin config using create-plugin sdk

aangelisc:


asimpson (commented):
`yarn build` works for me just fine but `yart test:ci` has 20 or failures with this error: 
```
Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

SyntaxError: Unexpected token 'export'
```

I'm not sure if this is related to the comment around "ESM errors with Jest" or not.

PR #610, opened 2023-08-22
`ConfigEditor` updates

aangelisc:
For standardisation across Grafana we're updating datasource config editors to be more consistent. This is a part of https://github.com/grafana/oss-plugin-partnerships/issues/225.

**This is a breaking change as the plugin now has a minimum Grafana dependency of 9.5.0**

The config editor has been updated to replace `InlineField` with `Field` and the config editor has been refactored to use the new config editor structural components. There is also now individual validation on required fields. I've also updated the `ConfigHelp` component so it will display by default on new datasources but not on ones that are already configured. Also fixed OpenAI api key prop name for consistency and bumped `@grafana/ui` package.

Node version has also been upgraded to 18.

image

Fixes grafana/oss-plugin-partnerships#232


src/components/ConfigEditor/ConfigHelp.tsx
```diff
@@ -1,10 +1,24 @@
...
+import React, { useMemo } from 'react';
+import { AdxDataSourceOptions, AdxDataSourceSecureOptions } from 'types';
+import { isCredentialsComplete } from './AzureCredentials';
 
-interface ConfigHelpProps {}
+interface ConfigHelpProps {
+  options: DataSourceSettings<AdxDataSourceOptions, AdxDataSourceSecureOptions>;
+}
+
+const ConfigHelp: React.FC<ConfigHelpProps> = ({ options }) => {
```
asimpson, line 11: I believe we're still trying to [avoid](https://github.com/grafana/grafana/blob/main/contribute/style-guides/frontend.md#how-to-declare-functional-components) using `React.FC`. So I think this should be something like this:

```suggestion
function ConfigHelp({ options }: ConfigHelpProps) {
```

src/components/ConfigEditor/index.tsx
```diff
@@ -60,12 +74,22 @@ const ConfigEditor: React.FC<ConfigEditorProps> = (props) => {
...
-      <ConfigHelp />
+    <>
+      <DataSourceDescription
+        dataSourceName="Azure Data Explorer"
+        docsLink="https://grafana.com/grafana/plugins/grafana-azure-data-explorer-datasource/"
+        hasRequiredFields
+      />
+      {/* TODO add dividers */
+      /* <Divider/>*/}
+      <ConfigHelp options={options} />
```
asimpson, line 85: I believe the `Divider` will help/fix this but right now when ConfigHelp is collapsed it stacks right on top of Connection Details.

image

asimpson (dismissed):
Looks good :art:

 I'd want to see it with the divider first

src/components/ConfigEditor/index.tsx
aangelisc: Yeah the `Divider` will definitely fix this 😊

src/components/ConfigEditor/ConfigHelp.tsx
aangelisc: To avoid too many changes would you be okay if I left this for a separate refactor? I think we're actually doing this for most of the functional components so I'd like to do them all at once in a clean PR, how does that sound?

asimpson: :+1: sounds good to me

bossinc (dismissed):

.config/README.md
```diff
@@ -100,21 +120,45 @@ const config = async (env): Promise<Configuration> => {
...
+  grafana:
+    container_name: 'myorg-basic-app'
+    build:
+      context: ./.config
+      args:
+        grafana_version: ${GRAFANA_VERSION:-9.1.2}
+        grafana_image: ${GRAFANA_IMAGE:-grafana}
+```
+
+In this example we are assigning the environment variable `GRAFANA_IMAGE` to the build arg `grafana_image` with a default value of `grafana`. This will give you the possibility to set the value while running the docker-compose commands which might be convinent in some scenarios.
```
bossinc, line 162:
```suggestion
In this example, we assign the environment variable `GRAFANA_IMAGE` to the build arg `grafana_image` with a default value of `grafana`. This will allow you to set the value while running the docker-compose commands, which might be convenient in some scenarios.
```

I don't think this line is needed, but here are a few suggestions if you want to keep the line.

.config/README.md
```diff
@@ -100,21 +120,45 @@ const config = async (env): Promise<Configuration> => {
...
 **Update for `dev`:**
+
 ```diff
 -"dev": "webpack -w -c ./.config/webpack/webpack.config.ts --env development",
 +"dev": "webpack -w -c ./webpack.config.ts --env development",
 ```
+
+### Configure grafana image to use when running docker
+
+By default `grafana-enterprise` will be used as the docker image for all docker related commands. If you want to override this behaviour simply alter the `docker-compose.yaml` by adding the following build arg `grafana_image`.
```
bossinc, line 145:
```suggestion
By default, `grafana-enterprise` will be used as the docker image for all docker related commands. If you want to override this behavior, simply alter the `docker-compose.yaml` by adding the following build arg `grafana_image`. 
```

.config/README.md
```diff
@@ -50,7 +52,23 @@ There are two configuration in the project root that belong to Jest: `jest-setup
 **`jest-setup.js`:** A file that is run before each test file in the suite is executed. We are using it to
 set up the Jest DOM for the testing library and to apply some polyfills. ([link to Jest docs](https://jestjs.io/docs/configuration#setupfilesafterenv-array))
 
-**`jest.config.js`:** The main Jest configuration file that is extending our basic Grafana-tailored setup. ([link to Jest docs](https://jestjs.io/docs/configuration))
+**`jest.config.js`:** The main Jest configuration file that extends the Grafana recommended setup. ([link to Jest docs](https://jestjs.io/docs/configuration))
+
+#### ESM errors with Jest
+
+A common issue found with the current jest config involves importing an npm package which only offers an ESM build. These packages cause jest to error with `SyntaxError: Cannot use import statement outside a module`. To work around this we provide a list of known packages to pass to the `[transformIgnorePatterns](https://jestjs.io/docs/configuration#transformignorepatterns-arraystring)` jest configuration property. If need be this can be extended in the following way:
```
bossinc, line 59:
```suggestion
A common issue with the current jest config involves importing an npm package that only offers an ESM build. These packages cause jest to error with `SyntaxError: Cannot use import statement outside a module`. To work around this, we provide a list of known packages to pass to the `[transformIgnorePatterns](https://jestjs.io/docs/configuration#transformignorepatterns-arraystring)` jest configuration property. If need be, this can be extended in the following way:
```

bossinc (dismissed):

.config/README.md
aangelisc: I'll also PR these changes to the upstream repo 😊

.config/webpack/utils.ts
```diff
@@ -1,42 +1,59 @@
...
+import os from 'os';
 import path from 'path';
 import util from 'util';
-import glob from 'glob';
+import { glob } from 'glob';
 import { SOURCE_DIR } from './constants';
 
-const globAsync = util.promisify(glob);
+
+export function isWSL() {
```
adamyeats, line 10: Could this function go in a shared lib? Seen this in a couple of places now, it's probably quite useful

.config/webpack/utils.ts
```diff
@@ -1,42 +1,59 @@
...
 
 export function getPackageJson() {
   return require(path.resolve(process.cwd(), 'package.json'));
 }
 
-export function getPluginId() {
-  const { id } = require(path.resolve(process.cwd(), `${SOURCE_DIR}/plugin.json`));
-
-  return id;
+export function getPluginJson() {
```
adamyeats, line 30: Like this, this makes much more sense given the companion function `getPackageJson` above.

.config/webpack/webpack.config.ts
```diff
@@ -12,188 +13,201 @@ import path from 'path';
...
-                dynamicImport: true,
+
+    context: path.join(process.cwd(), SOURCE_DIR),
+
+    devtool: env.production ? 'source-map' : 'eval-source-map',
+
+    entry: await getEntries(),
+
+    externals: [
+      'lodash',
```
adamyeats, line 37: Should we alphabetise these array values?

.config/webpack/webpack.config.ts
```diff
@@ -12,188 +13,201 @@ import path from 'path';
...
-    unsafeCache: true,
-  },
-});
+  }
+
+  if(isWSL()) {
+    baseConfig.watchOptions = {
+      poll: 3000,
+      ignored: /node_modules/,
+    }}
```
adamyeats, line 206: Should the `}` be propped to the next line?

.config/webpack/webpack.config.ts
```diff
@@ -12,188 +13,201 @@ import path from 'path';
...
+
+
+  return baseConfig;
+
```
adamyeats, line 210: Empty line here is probably not necessary

.github/workflows/detect-breaking-changes.yml
```diff
@@ -6,9 +6,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-node@v3
+      - uses: actions/setup-node@v3.8.1
```
adamyeats, line 9: Why the hard version lock here? I'm not sure if locking to a patch version is beneficial to us here.

src/components/ConfigEditor/Divider.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { Divider as GrafanaDivider, useTheme2 } from '@grafana/ui';
+import { config } from '@grafana/runtime';
+import { isVersionGtOrEq } from 'version';
+
+export function Divider() {
+  const theme = useTheme2();
+  return isVersionGtOrEq(config.buildInfo.version, '10.1.0') ? (
```
adamyeats, line 8: I think this reads better as just a regular `if`/`else` that returns early if `true`.

src/components/ConfigEditor/OpenAIConfig.tsx
```diff
@@ -35,25 +36,29 @@ export function OpenAIConfig(props: Props) {
...
-        </InlineField>
-        <InlineField>
-          <Button variant="secondary" type="button" onClick={onResetAPIKey}>
-            Reset
-          </Button>
-        </InlineField>
+        <Field
+          label="OpenAI API Key"
+          htmlFor="openai-api-key"
+          description="Supplying an OpenAI API key will allow users to make use of the generate queries feature."
```
adamyeats, line 44:
```suggestion
          description="Supplying an OpenAI API key will allow users to make use of the \"Generate queries\"  feature."
```

Sorry if I didn't escape this correctly! 😕

src/components/ConfigEditor/OpenAIConfig.tsx
```diff
@@ -35,25 +36,29 @@ export function OpenAIConfig(props: Props) {
...
+        >
+          <div className="width-30" style={{ display: 'flex', gap: '4px' }}>
+            <Input
+              id="openai-api-key"
+              aria-label="OpenAI API Key"
+              placeholder={secureJsonFields?.OpenAIAPIKey ? 'configured' : ''}
+              value={secureJsonData?.OpenAIAPIKey ?? ''}
+              onChange={onAPIKeyChange}
+              data-testid={'openai-api-key'}
```
adamyeats, line 53:
```suggestion
              data-testid="openai-api-key"
```

src/components/QueryEditor/index.tsx
```diff
@@ -9,5 +9,9 @@ import { AdxDataSourceOptions, KustoQuery } from 'types';
 type Props = QueryEditorProps<AdxDataSource, KustoQuery, AdxDataSourceOptions>;
 
 export const QueryEditor: React.FC<Props> = (props) => {
-  return config.featureToggles.adxLegacyEditor ? <LegacyQueryEditor {...props} /> : <NewQueryEditor {...props} />;
+  return (config.featureToggles as any).adxLegacyEditor ? (
```
adamyeats, line 12: I think this reads better as just a regular `if`/`else` that returns early if `true`.

adamyeats, line 12: Also, can you please leave a code comment to explain the use of `any` here? So that if anyone encounters this line in future, we can go back and check whether the issue preventing us from using `config.featureToggles` is solved.

.config/webpack/utils.ts
aangelisc: This is autogenerated based on `npx @grafana/create-plugin@latest update` so I guess technically already in a shared lib 😄

.config/webpack/webpack.config.ts
aangelisc: Also autogenerated 😊

.config/webpack/webpack.config.ts
aangelisc: I think the `eslint` config for the `create-plugin` tool may be slightly different to our config so I haven't touched these files

.config/webpack/webpack.config.ts
aangelisc: Same as above

.github/workflows/detect-breaking-changes.yml
aangelisc: Removed the pinning

src/components/QueryEditor/index.tsx
aangelisc: Added a comment

aangelisc: I've updated this, personally I'm more of an advocate for ternary operators over `if/else` as I find them simpler but maybe they're not as readable.

PR #616, opened 2023-09-05
Add Docker image builds and post `docker pull` commands for local testing

adamyeats:
Related to https://github.com/grafana/grafana/issues/73287.

This PR gives us the ability to just spin up a Docker image to test any changes made in a local instance of Grafana. When the CI pipeline is complete, the Github Actions bot will publish a helpful PR comment so that the reviewer can just copy/paste the command into their terminal.

The image is tagged with the long SHA of the most recent commit. At least in the short-term, all plugins will push to the same Docker Hub repo, so we use the long SHA to minimise the possibility of collision (it's also much easier to get that from the build context than the short SHA).

### DIFF
--- a/.github/workflows/pull-request-image.yml
+++ b/.github/workflows/pull-request-image.yml
@@ -1,0 +1,111 @@
+name: Create Docker image for local testing
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+    branches:
+      - main
+
+permissions:
+  pull-requests: write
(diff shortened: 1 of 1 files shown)


aangelisc (dismissed):
Relevant comment [here.](https://github.com/grafana/falconlogscale-datasource/pull/97#pullrequestreview-1611517173)

aangelisc (approved):

PR #618, opened 2023-09-11
Making Azure Data Explorer show up under the Logging category

baselsalam:


aangelisc:
Hi @baselsalam, thank you for the contribution. What's the reason for adding the logging category? I wouldn't say that ADX is specifically a logging datasource 🤔

baselsalam:
This would allow using ADX in the "TracesToLogs" feature, which currently
only shows data sources under the "Logging" category.

ADX seems like it should be a logging data source. I've been using it for
almost a decade and the primary use case is logging.
It's pitched as an alternative to Elasticsearch/Splunk, and I've helped
numerous teams transition from those to ADX.

aangelisc:
Hi @baselsalam,

> This would allow using ADX in the "TracesToLogs" feature, which currently
only shows data sources under the "Logging" category.

This is not the case, the supported datasources for `TracesToLogs` are defined in code. Setting the `logging` value for `category` will not display ADX datasources as an option for `TracesToLogs`. This would be a feature request that will require more thought/work. There is a discussion related to this [here](https://github.com/grafana/azure-data-explorer-datasource/discussions/620).

aangelisc:
I'm going to close this PR as the implementation isn't as straightforward as adding the `logging` property. Please continue the conversation in the linked discussion 😊

PR #699, opened 2023-11-08
OpenAI: Incorporate the LLM plugin

asimpson:
We will have to support our own OpenAI integration in addition to the new integration via the LLM plugin. This shouldn't be too arduous and gives us additional features like Azure OpenAI without additional code on our part. I am removing the Config option for OpenAI to drive new users to use the LLM plugin path.

I also fixed a few little bugs in regards to the query editor and generated queries which should make the overall experience smoother.

To verify:

1. Navigate to the OpenAI editor and note the Alert directing you to install the plugin.
2. Install the plugin and configure it with a valid API key.
3. Return to the OpenAI editor and note the Alert is gone and you are able to generate queries.


src/components/QueryEditor/OpenAIEditor.tsx
```diff
@@ -64,24 +72,52 @@ export const OpenAIEditor: React.FC<RawQueryEditorProps> = (props) => {
...
-    datasource
-      .generateQueryForOpenAI(`${baselinePrompt}${prompt}"""`)
-      .then((resp) => {
-        setWaiting(false);
-        setGeneratedQuery(resp);
-        onRawQueryChange(resp);
+    if (enabled) {
+      newGenerateQuery(); //eventually this will replace this function
+    } else {
+      datasource
```
aangelisc, line 82: Will this still work without the OpenAI config?

asimpson: Shoot, I thought it would work since the key was already saved but you're correct now that I double-check it. I'll add the config back but disable the field or something.

asimpson: I was right and the API being down was the problem :sweat_smile: . If users have configured an API token previously the removal of the Config component won't prohibit them from using it.

jtheory: :sob: OpenAI had an outage today -- we'd just deployed a new feature to ops and were testing, totally threw us off, too

PR #727, opened 2023-11-30
Cluster selector

bossinc:
Overview of change
-
- In settings, `Cluster url` was renamed to `Default cluster url`. The cluster url in settings is now the default option when a cluster is not selected in a query.
- The query editor has a select cluster drop down. You can select a cluster your Azure creds have access to

Code change notes
-
- Added a new resource call, `getClusters`.
- Changed `getSchema` and `getDatabase` calls to be POST to accept the cluster uri
- Added a new client to `httpClient` to `Client` to support auth scoping for ARG queries. I would prefer to have one client, but I could not make it work with the `grafana-azure-sdk-go`


pkg/azuredx/client/httpclient.go
```diff
@@ -14,13 +14,55 @@ import (
...
 	if err != nil {
 		return nil, err
 	}
 
+	authOpts, err := getAuthOpts(azureSettings, dsSettings, azureCloud)
+	if err != nil {
+		return nil, err
+	}
+
+	scopes, err := getAdxScopes(azureCloud, dsSettings.ClusterURL) // HELP! This is used to get China or US ADX. not sure how to test this stuff
```
bossinc, line 29: I am not sure how to test USA and China gov stuff :smiling_face_with_tear:

pkg/azuredx/client/client.go
```diff
@@ -18,23 +19,29 @@ import (
...
+	ARGClusterRequest(ctx context.Context, payload models.ARGRequestPayload, additionalHeaders map[string]string) ([]models.ClusterOption, error)
 }
 
 var _ AdxClient = new(Client) // validates interface conformance
 
 // Client is an http.Client used for API requests.
 type Client struct {
-	httpClient *http.Client
+	httpClientAzureCloud *http.Client
+	httpClientManagement *http.Client
```
bossinc, line 31: It would be nice to not have 2 clients, but I could not think of a way around this. Any ideas?

src/components/ConfigEditor/ConnectionConfig.tsx
```diff
@@ -16,11 +16,8 @@ const ConnectionConfig: React.FC<ConnectionConfigProps> = ({ options, updateJson
...
       <Field
-        label="Cluster URL"
-        description="The cluster url for your Azure Data Explorer database."
-        required
-        error={'Cluster URL is required'}
-        invalid={!options.jsonData.clusterUrl}
+        label="Default cluster url"
+        description="The default cluster url for your Azure Data Explorer database."
       >
         <Input
```
bossinc, line 22: This should be the drop down same drop down field as the query editor

aangelisc (commented):
Great stuff @bossinc! I've left some nits in the comments as well but we can discuss them if you like. I think this would probably benefit from some additional testing as well across the frontend and backend changes.

I experienced a bug that might be worth investigating before we merge:

- The default cluster URL doesn't always get set properly in the Explore or dashboard views.

Also, I'm wondering if, when retrieving the clusters, we could test if the user/identity actually has access to a cluster before displaying it to avoid them attempting to connect to resources they can see but don't actually have access to. Wdyt?

aangelisc:
Interesting edge case to make note of, if a cluster is stopped then it still shows up in the cluster list but isn't queryable. We should create a separate issue after this to tackle that so users don't unintentionally query clusters that aren't prepared.

pkg/azuredx/client/client.go
```diff
@@ -58,28 +82,65 @@ func (c *Client) TestRequest(ctx context.Context, datasourceSettings *models.Dat
...
+
+	resp, err := c.httpClientManagement.Do(req)
+	if err != nil {
+		return err
+	}
+
+	defer resp.Body.Close()
+
+	if resp.StatusCode == 403 {
+		return errors.New("The client does not have permission to use Azure Resource Graph. The cluster select in the query editor and the default cluster URL in this data source config will not be populated.")
```
aangelisc, line 125:
```suggestion
		return errors.New("The client does not have permission to use Azure Resource Graph. The cluster select in the query editor will not be populated.")
```

src/components/QueryEditor/QueryHeader.tsx
```diff
@@ -44,6 +45,8 @@ const adxTimeFormat: SelectableValue<string> = {
 export const QueryHeader = (props: QueryEditorHeaderProps) => {
   const { query, onChange, schema, datasource, dirty, setDirty, onRunQuery, templateVariableOptions } = props;
   const { rawMode, OpenAI } = query;
+  const cluster = query.clusterUri;
```
aangelisc, line 48: This should be a state variable I think. That would then allow you to conveniently update this in the `useEffect` if no cluster URI is specified.

src/components/QueryEditor/QueryHeader.tsx
```diff
@@ -115,6 +129,14 @@ export const QueryHeader = (props: QueryEditorHeaderProps) => {
           setShowWarning(false);
         }}
       ></ConfirmModal>
+      <InlineSelect
+        label="Cluster"
+        aria-label="Cluster"
+        options={clusters}
```
aangelisc, line 135: We should add template variable options here I think 😊

aangelisc (approved):
Just a couple more nits but otherwise LGTM! Great work on this 💯

pkg/azuredx/client/client.go
```diff
@@ -18,27 +20,49 @@ import (
...
+	httpClientManagement, err := newHttpClientManagement(ctx, instanceSettings, dsSettings, azureSettings, credentials)
+	if err != nil {
+		return nil, err
+	}
+	return &Client{httpClientKusto: httpClientAzureCloud, httpClientManagement: httpClientManagement}, nil
 }
 
 // TestRequest handles a data source test request in Grafana's Datasource configuration UI.
 func (c *Client) TestRequest(ctx context.Context, datasourceSettings *models.DatasourceSettings, properties *models.Properties, additionalHeaders map[string]string) error {
+	if err := c.testManagementClientAndGetCluster(ctx, datasourceSettings, additionalHeaders); err != nil {
```
aangelisc, line 50: One final nit from me 🙃 I think we should run this test but not fail the health check if it doesn't succeed. This test should just serve as a warning that certain functionality will not work but it's not an error.

pkg/azuredx/client/client.go
```diff
@@ -18,27 +20,49 @@ import (
...
+		return err
+	}
+
+	if datasourceSettings.ClusterURL == "" {
+		return nil
+	}
+
+	if err := c.testKustoClientWithDefaultCluster(ctx, datasourceSettings, datasourceSettings.ClusterURL, properties, additionalHeaders); err != nil {
```
aangelisc, line 58: One other final nit 😜 what happens here if there is no `ClusterURL` now that it's an optional property

bossinc:
I can't get the e2e tests to work. I've tried debugging locally but it isn't finding Grafana aria labels and the drone tests do not say where they are failing.

pkg/azuredx/datasource.go
```diff
@@ -95,14 +95,23 @@ func (adx *AzureDataExplorer) CallResource(ctx context.Context, req *backend.Cal
...
 			Status:  backend.HealthStatusError,
 			Message: err.Error(),
 		}, nil
 	}
 
+	err = adx.client.TestARGsRequest(ctx, adx.settings, models.NewConnectionProperties(adx.settings, nil), headers)
+	if err != nil {
+		return &backend.CheckHealthResult{
+			Status:  backend.HealthStatusOk,
+			Message: "Success connecting to Azure Data Explore, but unable to connect to Azure Resource Graph to get clusters: " + err.Error(),
```
aangelisc, line 111:
```suggestion
			Message: "Success connecting to Azure Data Explorer, but unable to connect to Azure Resource Graph to retrieve clusters: " + err.Error(),
```

aangelisc (approved):
Amazing work @bossinc! This is a great feature that will deliver so much value 😄

PR #768, opened 2024-01-29
Enable PDC for ADX

leandro-deveikis:
### Background
This PR enables support for [the secure socks proxy](https://grafana.com/docs/grafana/next/setup-grafana/configure-grafana/proxy/), which is a new feature in Grafana 10. The feature allows datasource connections to be proxied through a socks5 proxy with TLS. This enables Grafana users to connect to datasources that live in different networks than where Grafana is running when they cannot open up ports to the public internet.

### Implementation notes
- The feature itself no longer sits behind a feature toggle in Grafana OSS, but a feature toggle is being used in all non-core datasources until the datasource can upgrade the frontend dependency@grafana/runtime to at least 10.0.0. At that point, we can switch [this line](https://github.com/grafana/azure-data-explorer-datasource/pull/768/files#diff-46e37405ec7a5456cabd314e64f20aa4bc69ed018f57d8cab41ee85c2abb7c3cR121) to look at config.secureSocksDSProxyEnabled, rather than the feature toggle.
- As ADX datasource uses httpclient from grafana-plugin-sdk, there are not mayor changes needed in the backend, as the configuration is read and set in the SDK constructor functions for the httpclient.

### Dependency updates:
- Not needed, plugin-sdk was already updated to the required version. 

### Tests:
- This was tested locally, running the PDC server and ensuring the connections went through the PDC tunnel. 

Screenshot of configuration page (running locally):
image

### DIFF
--- a/cspell.config.json
+++ b/cspell.config.json
@@ -161,6 +161,7 @@
     "Ymax",
     "Ymin",
     "llms",
+    "proxying",
     ""
   ]
 }
--- a/src/components/ConfigEditor/index.tsx
+++ b/src/components/ConfigEditor/index.tsx
@@ -1,6 +1,8 @@
 import React, { useCallback, useEffect, useMemo } from 'react';
-import { DataSourcePluginOptionsEditorProps } from '@grafana/data';
+import { FeatureToggles, DataSourcePluginOptionsEditorProps } from '@grafana/data';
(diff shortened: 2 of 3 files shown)


bossinc (approved):
:bow: thank you!

PR #769, opened 2024-01-30
Log support upgrade

bossinc:
- Changed the table test to use :sparkles:CheckGoldenJSONFrame:sparkles: 
- Marked as a data source that supports logs
- Highlights search terms. When KQL has `| where someStringField = 'value'` value will be highlighted in the log line
- Log lines will be color coded depending on the log level
- Filter out labels


asimpson (approved):
:art: so nice!

PR #781, opened 2024-02-08
Remove json fields as this breaks test because of ordering

aangelisc:
The logs table frame test used the same data as the traces test. This is problematic as there's no guarantee the ordering in the slice when parsed will be consistent (which is why tests are intermittently failing). For traces, we order the slice but we aren't doing this for logs at the moment. Until we decide whether or not we should do this I've simplified the mock data to not include JSON objects.

### DIFF
--- a/pkg/azuredx/models/table_test.go
+++ b/pkg/azuredx/models/table_test.go
@@ -75,7 +75,7 @@ func TestResponseToFrames(t *testing.T) {
 		},
 		{
 			name:     "logs should be converted to dataframe appropriately",
-			testFile: "adx_traces_table.json",
+			testFile: "adx_logs_table.json",
 			errorIs:  assert.NoError,
 			format:   data.VisTypeLogs,
 		},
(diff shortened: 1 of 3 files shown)


pkg/azuredx/models/table_test.go
```diff
@@ -93,7 +93,7 @@ func TestResponseToFrames(t *testing.T) {
 				return
 			}
 
-			experimental.CheckGoldenJSONFrame(t, "./testdata/golden_data", tt.name, frames[0], false)
+			experimental.CheckGoldenJSONFrame(t, "./testdata/golden_data", tt.name, frames[0], true)
```
asimpson, line 96: Do we want `updateFile` arg to be `true`? Or should it remain `false`?

asimpson (commented):
Looks good :trophy: 

Just one question...

bossinc (approved):

asimpson (approved):

adamyeats (approved):

PR #815, opened 2024-03-15
Bypass trusted endpoint enforcement for cluster request

### LINKED ISSUE #814: Unable to load clusters when trusted endpoint enforcement enabled
**What happened**:

When `enforce_trusted_endpoints` enabled, it's not possible to use the cluster selection in ADX datasource.

The call to Azure ARM API should bypass the trusted endpoints enforcement since the endpoint isn't configurable by user.

Enable endpoint enforcement in Grafana config:

```
[plugin.grafana-azure-data-explorer-datasource]
enforce_trusted_endpoints = true
```

Test datasource or try create a new panel using the datasouce.

**Screenshots**

image

**Environment**:

- Grafana version: any
- Plugin version: 4.9.0
- OS Grafana is installed on:
- User OS & Browser:
- Others:

kostrse:
Azure Management API calls do not need to have trusted endpoint enforcement because the endpoint isn't supplied by the user.

It fixes bug which prevents loading of clusters when trusted endpoint enforcement is enabled.

Fixes #814

### DIFF
--- a/pkg/azuredx/client/httpclient.go
+++ b/pkg/azuredx/client/httpclient.go
@@ -21,7 +21,7 @@ func newHttpClientAzureCloud(ctx context.Context, instanceSettings *backend.Data
 		return nil, err
 	}
 
-	authOpts, err := getAuthOpts(azureSettings, dsSettings, azureCloud)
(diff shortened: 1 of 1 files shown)


aangelisc (approved):
Thank you @kostrse! Nice fix 😊

PR #827, opened 2024-03-22
Prepare 5.0.0

aangelisc:
- Bump version
- Update CHANGELOG

This is a major release as we've removed the `LegacyQueryEditor`


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "grafana-azure-data-explorer-datasource",
-  "version": "4.10.0",
+  "version": "4.11.0",
```
adamyeats, line 3: This doesn't match `5.0.0`.

adamyeats (changes requested):

package.json
aangelisc: Oops, thanks for catching!

adamyeats (dismissed):

bossinc (approved):

PR #835, opened 2024-03-28
Do not execute a query on editor load

### LINKED ISSUE #796: When creating new Annotation using Azure Data Explorer as datasource, automatically loads all records from the first table causing out of memory failures
# What happened?
When I create new annotation, and select Azure Data Explorer (Kusto) as data source, the query preview window automatically loads the first database and first table and runs a query on it. This means it loads all the records without any limitations exceeding maximum memory allowed if table is large enough.

# What did you expect to happen?
When creating new annotations, by default the first load shouldn't be automatic, the user has to click "Test" to trigger it, this is how it works on regular panels right now.

Also it probably makes sense to always have a "| limit XXX" set on all these queries by default to prevent this from causing so much damage.

# Did this work before?
don't know, haven't used this before.

# How do we reproduce it?
Go to any dashboard
Open Annotations settings
Click "New Query"
Select "Azure Data Explorer" datasource from data source selector
Wait a minute
An exception will appear:
Query execution has exceeded the allowed limits (80DA0001): 'sort' operator has exceeded the memory budget (5368709120) during evaluation. Results may be incorrect or incomplete (E_RUNAWAY_QUERY; see https://aka.ms/kustoquerylimits).. [0]Kusto.Data.Exceptions.KustoServicePartialQueryFailureLimitsExceededException: Query execution has exceeded the allowed limits (80DA0001): 'sort' operator has exceeded the memory budget (5368709120) during evaluation. Results may be incorrect or incomplete (E_RUNAWAY_QUERY; see https://aka.ms/kustoquerylimits)

# Environment (with versions)?
Grafana: 9.5.15
OS: Azure Managed Grafana
Browser: Edge

Grafana platform?
Other

Datasource(s)?
Azure Data Explorer

NOTE: This is posted here as well - https://github.com/grafana/grafana/issues/82876

aangelisc:
When opening the `VisualQueryEditor` the current behaviour is to auto-select the first table and all of its columns, leading to a query being immediately run.

This is not the best behaviour as these queries could lead to large amounts of data being returned. This is also not the standard Grafana behaviour.

This PR changes the behaviour to not auto-run a query by not auto-selecting a table.

Fixes #796

### DIFF
--- a/cypress/integration/smoke.spec.ts
+++ b/cypress/integration/smoke.spec.ts
@@ -32,7 +32,7 @@ const addAdxVariable = (
   name: string,
   type: AdxQueryType,
   isFirst: boolean,
-  options?: { cluster?: string, database?: string; table?: string }
+  options?: { cluster?: string; database?: string; table?: string }
 ) => {
   if (isFirst) {
     e2e.components.PageToolbar.item('Dashboard settings').click();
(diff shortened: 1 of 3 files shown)


bossinc (approved):
Good call out

PR #849, opened 2024-04-09
Allow additional trusted hosts

matrujillo98:
Add an option to support additional trusted endpoints. This will allow the users to trust specific hosts, and not need to drop the whole validation of endpoints.

I tested it locally, but unfortunately did not have proper proxy to fully test. Nonetheless, I used a local DNS forwarding to verify, but was not able to fully check due to the TLS certs.

image

### DIFF
--- a/pkg/azuredx/client/endpoints.go
+++ b/pkg/azuredx/client/endpoints.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/grafana/grafana-azure-sdk-go/azsettings"
 )
@@ -53,10 +54,15 @@ var azureDataExplorerEndpoints = map[string][]string{
 	},
 }
 
-func getAdxEndpoints(azureCloud string) ([]string, error) {
+func getAdxEndpoints(azureCloud string, trustedClustersURLs string) ([]string, error) {
 	if endpoints, ok := azureDataExplorerEndpoints[azureCloud]; !ok {
 		return nil, fmt.Errorf("the Azure cloud '%s' not supported by Azure Data Explorer datasource", azureCloud)
 	} else {
+		// Append the trusted URLs for all clouds
+		var trustedUrls = strings.Split(trustedClustersURLs, ",")
+		if len(trustedUrls) > 1 || trustedUrls[0] != "" {
+			endpoints = append(endpoints, trustedUrls...)
+		}
 		return endpoints, nil
 	}
 }
--- a/pkg/azuredx/client/endpoints_test.go
+++ b/pkg/azuredx/client/endpoints_test.go
@@ -1,0 +1,151 @@
+package client
+
+import (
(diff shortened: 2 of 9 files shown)


aangelisc:
Hi @matrujillo98! Thank you for the contribution! Before I review this I'd like to understand the use-case here. Is this to have a proxy in front of an ADX cluster with a custom DNS or something else?

matrujillo98:
Hey @aangelisc, so Kusto natively does not support failing over other clusters, thus you could use proxies on top of it with a traffic manager to allow such feature and have BCDR. Additionally, when talking about national clouds, it is not possible for a dev to query a Kusto cluster unless it has a national cloud identity which is not necessarily the case in all situations. Thus, you could use a proxy to centralize the access, and only allow grafana identity to get the information.

matrujillo98:
Btw, compared with a running grafana instance and the result is indeed an issue with validation before TLS:

image

aangelisc:
Hi @matrujillo98, apologies for the delay and thank you for the additional context! Unfortunately, in order to implement this we'd require some additional changes (that would require us to potentially rework some aspects of the `grafana-azure-sdk` or another area of Grafana.

We'd need to ensure that we only display the component for specifying additional trusted hosts when the trusted endpoints setting is enabled to avoid confusion. Additionally, we'd also need to add some config to allow server admins to disable this feature entirely for security reasons.

matrujillo98:
Hey @aangelisc, thanks for the response!

Sure, I could add such a flag for admins to block the feature, and make sure this only appears if the overall enforcement is in place.

About this:

> Unfortunately, in order to implement this we'd require some additional changes (that would require us to potentially rework some aspects of the grafana-azure-sdk or another area of Grafana.

I'm not quite sure why it would be the case, but I may be able to confirm e2e that this could work with current code if you'd like.

aangelisc:
Hi @matrujillo98,

Thank you for offering to take a look at this! The reason is the way that configuration gets passed from the Grafana instance to the plugin at the moment is via environment variables but this isn't sustainable and going forward we'd like to change this.  I will investigate how that can be done and if it looks like it'll be too long of a timeframe then I'll figure out an alternative method in order to get this PR in and unblock you 😊

aangelisc:
Hi @matrujillo98,

I haven't forgotten about this work! Would you be comfortable creating a PR in Grafana to add the additional configuration variable required to allow Grafana admins to enable/disable this functionality?

Thanks!

aangelisc:
Hi @matrujillo98! So, on rethinking this, I think we should make it so that the trusted URLs can only be specified by the server admin (to avoid someone modifying a data source and trusting an endpoint that the server admin may not want to trust. Are you okay to modify this PR to support this use-case or would you prefer if I did so?

aangelisc:
Closing this for now but I've created #1227 so that we can follow up on this work. Thank you @matrujillo98.

PR #800, opened 2024-02-21
Clouds from Azure React SDK

kostrse:
Use `@grafana/azure-sdk` to get list of available Azure clouds rather than a hardcoded list in the datasource itself.

PR #854, opened 2024-04-11
Clouds from Azure React SDK

aangelisc:
Use `@grafana/azure-sdk` to get list of available Azure clouds rather than a hardcoded list in the datasource itself.

Copied from #800


bossinc (approved):
Nice refactor!

PR #855, opened 2024-04-11
Fix template variables multi-select

### LINKED ISSUE #823: Template variables in Builder
**What happened**:  When using the query builder in the ADX plugin, the template variable of the dashboard cannot be used (like mentioned in the [note](https://github.com/grafana/azure-data-explorer-datasource?tab=readme-ov-file#templating-with-variables)) because it parses them as a unique string and not as a list of values.

**What you expected to happen**: Use the template variable also in the builder for a better user experience for users who are not confident with KQL.

1. Create a template variable for the dashboard
2. Populate it with some values by which you want to filter in the query
3. In the ADX Plugin Builder try to use a filter function based on that variable
4. Run the query

You can see that the clause is translated like  ` | where origin in ('$origin')`  (adding single quotes around the variable) instead of  `| where origin in ($origin)`.

I don't see this as a feature request (even if expected by the documentation) but as a bug, because the Builder itself is the feature for the plugin but it is not working as expected.

**Screenshots**

**Environment**:

- Grafana version: 10.3.4
- Plugin version: 4.9.0
- OS Grafana is installed on:  Azure Managed Grafana
- User OS & Browser:
- Others:

aangelisc:
Allow multi-select of template variables in the builder editor.

The problem here was that we were quoting the template variable values. I've tested this with multiple filters/group bys and the quoting doesn't seem necessary so I've removed it.

Let me know if you find any cases where this may not be the case!

Updated the docs to reflect that these variables will now work in the builder.

Also, a minor change to ensure there is a `table` selected in the query editor before we display the error asserting that there is no table schema.

Fixes #823


bossinc (commented):
This doesn't seem to work when I create constants without quotes, but I don't think we should be modifying template variable values in components. That should be done in applyTemplateVars function in datasource

adamyeats (dismissed):

javisst:
@aangelisc Do you have an update on this?

aangelisc:
Hi @javisst, not yet - I'm hoping to get to this next week.

bossinc (approved):

bossinc (dismissed):

bossinc (approved):

PR #862, opened 2024-04-16
Move tracking headers to client

### LINKED ISSUE #526: Move usage tracking from business layer to ADX client
[`modelQuery()`](https://github.com/grafana/azure-data-explorer-datasource/blob/f7e800acb971c53ce518226765a5ae3456f6e835/pkg/azuredx/datasource.go#L158) func takes `backend.User` as an argument and sets headers `x-ms-user-id`, `x-ms-client-request-id`.

This should not be a part of business layer and moved to the ADX client.

ADX client will have access to `adxcontext.CurrentUserContext` (#525) which includes `backend.User` to implement all necessary usage tracking transparently for business layer.

aangelisc:
As the `User` is now accessible via context the creation of the tracking headers can be done in the client rather than in the query logic.

Fixes #526

### DIFF
--- a/pkg/azuredx/resource_handler.go
+++ b/pkg/azuredx/resource_handler.go
@@ -141,8 +141,8 @@ func (adx *AzureDataExplorer) getSchema(rw http.ResponseWriter, req *http.Reques
 		QuerySource: "schema",
 	}
 
-	headers := map[string]string{}
-	response, err := adx.client.KustoRequest(req.Context(), cluster.ClusterUri, ManagementApiPath, payload, headers)
+	// Default to not sending the user request headers for schema requests
+	response, err := adx.client.KustoRequest(req.Context(), cluster.ClusterUri, ManagementApiPath, payload, false)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/azuredx/resource_handler.go
```diff
@@ -141,8 +141,8 @@ func (adx *AzureDataExplorer) getSchema(rw http.ResponseWriter, req *http.Reques
 		QuerySource: "schema",
 	}
 
-	headers := map[string]string{}
-	response, err := adx.client.KustoRequest(req.Context(), cluster.ClusterUri, ManagementApiPath, payload, headers)
+	// Default to not sending the user request headers for schema requests
```
bossinc, line 144: I'm not sure what this comment means. Are we just saying that we are not sending the user request header because it's a schema request? i think im being dense 😆

bossinc (approved):
I want to understand more about these tracking headers but everything seems to work!

pkg/azuredx/resource_handler.go
aangelisc: Exactly that, we don't need to send the header on every request. IIRC it should only be used for queries

PR #868, opened 2024-04-19
Use GetCloud from grafana-azure-sdk-go to get resource graph scope and fallback to using clusterUrl for ADX scope for unknown clouds

JonCole:
Using GetCloud from the azure sdk package gives us a consolidated location to get information about a specific cloud, allowing easier expansion of supported clouds in the future.  Also, even though Azure Public and Azure China use support special scopes, using the cluster url in the scope is the recommended practice for other clouds.


pkg/azuredx/client/scopes.go
```diff
@@ -2,43 +2,56 @@ package client
...
-		err := fmt.Errorf("the Azure cloud '%s' not supported by Azure Data Explorer datasource", azureCloud)
-		return nil, err
-	} else {
-		// TODO: #356 Generalize generation of scopes for all clouds
-		clusterUrl = strings.TrimSuffix(clusterUrl, "/")
-		scopes := []string{strings.Replace(scopeTmpl, "{clusterUrl}", clusterUrl, 1)}
-		return scopes, nil
+	scopeTmpl, ok := "", false
+	if scopeTmpl, ok = adxScopes[azureCloud]; !ok {
+		scopeTmpl = "{clusterUrl}/.default"
```
aangelisc, line 23: I'd add a clarifying comment here that this is applicable for all other clouds

aangelisc (approved):
LGTM, just one nit in the comments to address 😊 there's also a merge conflict that needs resolving

JonCole:
@aangelisc thanks for the review.  I have fixed the merge conflict and added the comment - please take another look.

aangelisc (approved):
Thanks @JonCole!

PR #880, opened 2024-05-08
Migrate to `plugin-e2e`

### LINKED ISSUE #844: Replace grafana/e2e testing with plugin-e2e
Migrating away from Cypress to `plugin-e2e` (which uses Playwright). There are multiple reasons for doing this:

- `grafana/e2e` is deprecated and no longer being released.
- We've had numerous issues with Cypress and as we have an E2E package dedicated to plugins that is being maintained this should be far more reliable.
- We can now test against multiple Grafana versions OOTB which is awesome.

aangelisc:
Migrate the E2E tests from Cypress to Playwright and switch to using the `plugin-e2e` package instead of `@grafana/e2e`.

I've tried to migrate all the existing tests over but there are some gaps:

- The dashboard importing test doesn't work but I'm not sure how much value this adds so I've removed it for now
- The original variable test that we had cannot be implemented in the same way using `plugin-e2e` so I've marked it to be skipped. We'll have to do something like create a dashboard with the template variables specified and then select the appropriate variable values. We can also use this to test each variable individually I think.

I've removed the `provisioning` folder from the git ignore so that we can include examples of how to provision the data source in the repo. This file is also used to provision the data source for E2E tests.

The secrets are supplied from Vault as environment variables for the tests.

I've also included some basic smoke tests that can run from forks.

I've updated dependencies using the `create-plugin` tool.

Finally, I fixed a little bug that I found when building the E2E tests where the code editor wouldn't show properly if the schema is empty.

Closes #844

Depends on grafana/plugin-tools#942


adamyeats:
It looks like the templating doesn't work. I see the name `${{ matrix.GRAFANA_IMAGE.name }}@${{ matrix.GRAFANA_IMAGE.VERSION }}` as a skipped check above. Is that something that will go away once merged maybe?

aangelisc:
> It looks like the templating doesn't work. I see the name `${{ matrix.GRAFANA_IMAGE.name }}@${{ matrix.GRAFANA_IMAGE.VERSION }}` as a skipped check above. Is that something that will go away once merged maybe?

No idea tbh, the template works correctly for the other versions so maybe it's a quirk or because it's skipping the fork e2e tests instead.

adamyeats (approved):

PR #913, opened 2024-06-10
Update azure-sdk package reference to 0.0.3

JonCole:
This brings in an important fix in that package


adamyeats (approved):

adamyeats:
Thanks @JonCole!

PR #921, opened 2024-06-13
Load custom clouds from context and remove hard coded URLs

JonCole:
This change fixes a problem where the ADX Plugin doesn't load the custom clouds list into the backend from grafana.  It also removes some hard coded URLs and uses the ones from the custom cloud list instead.


bossinc (approved):
Looks good! Thank you!

JonCole:
Thanks @bossinc  and @adamyeats !

PR #954, opened 2024-07-15
Add context to the user when Kusto returns an error

voidbar:
Hi, 
Currently if a user writes a malformed Kusto query, or the Kusto service is having a bad time with the query the user sends - the error you get back is a bit meaningless, and doesn't help the user resolve its issue:
image

I added the error's short description and type to the `Error` struct so we could propagate the error properly to show something like that (please note that the final formatting in the PR is a bit different):
image

Thanks!


asimpson (approved):
LGTM, thanks! :trophy:

PR #963, opened 2024-07-24
Sanitise URL when creating a new Kusto API request

adamyeats:
This PR adds logic to sanitise cluster URLs when creating a `KustoRequest`.

### DIFF
--- a/pkg/azuredx/client/client.go
+++ b/pkg/azuredx/client/client.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"math/rand"
 	"net/http"
+	"net/url"
 
 	"github.com/grafana/grafana-azure-sdk-go/v2/azcredentials"
 	"github.com/grafana/grafana-azure-sdk-go/v2/azsettings"
@@ -17,6 +18,7 @@ import (
 	// 100% compatible drop-in replacement of "encoding/json"
 	json "github.com/json-iterator/go"
 
+	"github.com/grafana/azure-data-explorer-datasource/pkg/azuredx/helpers"
 	"github.com/grafana/azure-data-explorer-datasource/pkg/azuredx/models"
 )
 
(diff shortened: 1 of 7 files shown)


pkg/azuredx/resource_handler.go
```diff
@@ -252,3 +260,27 @@ func respondWithError(rw http.ResponseWriter, code int, message string, err erro
 		rw.WriteHeader(http.StatusInternalServerError)
 	}
 }
+
+// sanitizeClusterUri ensures the URI does not contain a query or fragment part
+func sanitizeClusterUri(clusterUri string) (string, error) {
+	// check for trailing question mark before parsing
+	if strings.HasSuffix(clusterUri, "?") {
+		return "", errors.New("clusterUri contains invalid query characters")
```
alyssabull, line 268: nit - is it worth making this error more specific "clusterUri contains invalid question mark" or something along those lines to differentiate from the error below?

adamyeats: Usually you'd be 100% right here, but in this case it actually makes more sense (IMO) to be unspecific, so as not to give potential attackers too much feedback about what they're doing wrong in the API response, and thus making it easier for them. We can see the line the stack trace has come from on our side, which should be good enough for debugging purposes.

aangelisc (dismissed):
LGTM, awesome work @adamyeats - tested locally and saw the appropriate errors 😊

pkg/azuredx/client/client.go
```diff
@@ -141,7 +153,12 @@ func (c *Client) testManagementClient(ctx context.Context, datasourceSettings *m
 		return fmt.Errorf("the Azure cloud '%s' doesn't have the required property for 'resourceManager'", c.cloudSettings.Name)
 	}
 
-	req, err := http.NewRequestWithContext(ctx, http.MethodPost, resourceManager+"/providers/Microsoft.ResourceGraph/resources?api-version=2021-03-01", bytes.NewReader(buf))
+	fullUrl, err := url.JoinPath(resourceManager, "/providers/Microsoft.ResourceGraph/resources?api-version=2021-03-01")
```
aangelisc, line 156: This will have to be what it was before as `JoinPath` doesn't correctly handle the `api-version` query param. Alternatively, the URL could be constructed and the param could be set on the `Query` property

pkg/azuredx/client/client.go
```diff
@@ -232,7 +254,12 @@ func (c *Client) ARGClusterRequest(ctx context.Context, payload models.ARGReques
 		return nil, fmt.Errorf("the Azure cloud '%s' doesn't have the required property for 'resourceManager'", c.cloudSettings.Name)
 	}
 
-	req, err := http.NewRequestWithContext(ctx, http.MethodPost, resourceManager+"/providers/Microsoft.ResourceGraph/resources?api-version=2021-03-01", bytes.NewReader(buf))
+	fullUrl, err := url.JoinPath(resourceManager, "/providers/Microsoft.ResourceGraph/resources?api-version=2021-03-01")
```
aangelisc, line 257: Same here 😊

aangelisc (approved):

PR #964, opened 2024-07-25
Improve accounting for fragments

aangelisc:
URL fragments were potentially missed, this PR accounts for them in the same manner as query params.


adamyeats (approved):
🏆

PR #979, opened 2024-08-09
Downgrade monaco-editor and exclude from dependabot

aangelisc:
We exclude `monaco-editor` from dependabot as it needs to be tied to the version included by `grafana/ui` to avoid compilation issues.


alyssabull (dismissed):
Nice 💯 LGTM

alyssabull (approved):
Thanks for fixing the frontend tests!

PR #1030, opened 2024-09-20
Improve sources for errors

aangelisc:
Some errors were assigned an incorrect source. I've updated these now.

Fixes grafana/data-sources#78

### DIFF
--- a/pkg/azuredx/client/client.go
+++ b/pkg/azuredx/client/client.go
@@ -230,7 +230,7 @@ func (c *Client) KustoRequest(ctx context.Context, clusterUrl string, path strin
 
 	resp, err := c.httpClientKusto.Do(req)
 	if err != nil {
-		return nil, err
+		return nil, errorsource.DownstreamError(err, false)
 	}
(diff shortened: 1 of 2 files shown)


njvrzm (dismissed):
Looks good 👍🏻

I do wonder if any of the errors higher up in `KustoRequest` (client.go lines 202, 207, 212) or `ARGClusterRequest` (client.go lines 258, 263, 268) should be labeled as Downstream...

aangelisc:
> Looks good 👍🏻
> 
> I do wonder if any of the errors higher up in `KustoRequest` (client.go lines 202, 207, 212) or `ARGClusterRequest` (client.go lines 258, 263, 268) should be labeled as Downstream...

I would consider 212 to be a plugin error but I think you may be right about the other two. 263 could be either a plugin error or a downstream error depending on your PoV 😄 Will push some changes!

njvrzm (approved):
🚀

bossinc (approved):

PR #1052, opened 2024-10-14
fix: migrate to @grafana/llm package for OpenAI functionality

### LINKED ISSUE #1051: Uncaught TypeError: self.crypto.randomUUID is not a function when trying to generate a query using openAI
## Environment Details 
Installed last version of Grafana(11.2.0)
Installed last version of LLM plugin(0.10.9)
Installed last version of Azure Data Explorer data source for Grafana (5.0.7)
Configured the plugin with Azure OpenAI
with 2 Models Base / Large
Image

How to reproduce the problem:
1. Create a Dashboard 
2. Create a Panel 
3. Use Azure Data Explorer Database as a Data Source
4. Use the OpenAI option to generate a query 
Image
5. Write something like "Get me 5 samples of the data" and click on **Generate query**
6. A loading icon will appear inside the **Generate query** Button but nothing happens. 
Image
7. The console logs show the following error:
Image

sd2k:
The @grafana/experimental package is being deprecated so development
of LLM-related functionality has been moved to a separate package.

The API is exactly the same so this _should_ just work.

Fixes #1051.

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,6 +99,7 @@
     "@grafana/azure-sdk": "^0.0.5",
     "@grafana/data": "10.4.1",
     "@grafana/experimental": "^1.7.0",
+    "@grafana/llm": "^0.11.0",
     "@grafana/runtime": "10.4.1",
     "@grafana/schema": "10.4.1",
     "@grafana/ui": "10.4.1",
--- a/src/components/QueryEditor/OpenAIEditor.tsx
+++ b/src/components/QueryEditor/OpenAIEditor.tsx
@@ -1,5 +1,5 @@
 import { GrafanaTheme2, QueryEditorProps, SelectableValue } from '@grafana/data';
-import { llms } from '@grafana/experimental';
+import { openai } from '@grafana/llm';
 import { getTemplateSrv, reportInteraction } from '@grafana/runtime';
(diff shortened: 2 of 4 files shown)


sd2k:
I've had to fix the resolution of some Grafana packages which were otherwise being duplicated because the @grafana/llm package has quite a loose dependency range; without this, we had both version 10.4.1 and 11.3.0 of the various Grafana packages (and all of their subdependencies) which was causing tests to fail.

aangelisc:
@sd2k, I'm not overly keen on adding in the resolutions. Can the dependencies for the LLM package be made stricter? Alternatively, we can look at updating the `grafana/*` dependencies in this package to 11+

sd2k:
Yeah I'm not really keen on it either, I was surprised that the relaxed restrictions caused the resolver to choose two separate versions rather than preferring a common version :confused:

Upgrading to 11+ would probably do the trick and feels like the right call, especially given those packages are injected by Grafana anyway.

sd2k:
This might be superseded by #1168 if/when that gets merged first.

aangelisc:
Hey @sd2k, this will be superseded. I'm going to close this one out but appreciate you taking a look at this!

PR #1180, opened 2025-02-07
Docs: Updated current-user-auth.md in light of auth changes within Grafana

naizerjohn-ms:
Hi @aangelisc,

Here is a documentation update for the current user authentication doc, referencing the recent changes to how this works within the SDK.

### DIFF
--- a/doc/current-user-auth.md
+++ b/doc/current-user-auth.md
@@ -19,7 +19,7 @@ enabled = true
 
 auth_url = https://login.microsoftonline.com/fd719c11-a91c-40fd-8379-1e6cd3c59568//oauth2/v2.0/authorize
 token_url = https://login.microsoftonline.com/fd719c11-a91c-40fd-8379-1e6cd3c59568/oauth2/v2.0/token
-client_authentication = client_secret_post
+client_authentication = # defaults to 'client_secret_post' (client secret)
 client_id = f85aa887-490d-4fac-9306-9b99ad0aa31d
 client_secret = 87808761-ff7b-492e-bb0d-5de2437ffa55
 managed_identity_client_id =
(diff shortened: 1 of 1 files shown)


doc/current-user-auth.md
```diff
@@ -7,20 +7,23 @@ in the future.
...
-Assuming that Azure AD authentication has been already configured, for example:
+Assuming that Azure AD authentication has been already configured, for example, configuring Azure AD authentication using client secret:
 
 ```ini
 [auth.azuread]
 enabled = true
 
 auth_url = https://login.microsoftonline.com/fd719c11-a91c-40fd-8379-1e6cd3c59568//oauth2/v2.0/authorize
 token_url = https://login.microsoftonline.com/fd719c11-a91c-40fd-8379-1e6cd3c59568/oauth2/v2.0/token
+client_authentication = client_secret_post
```
aangelisc, line 22: Just a thought - as this is the default, we can exclude this line I believe. It may be confusing for existing users who will think they need to add this even when the user-auth should still function as expected right?

naizerjohn-ms: Yes! I'll remove the set value of `client_authentication`, but I think readers should know that it can be set if they want to override the default.

naizerjohn-ms:
@aangelisc check out the recent commit! Hopefully this clears up your point?

naizerjohn-ms:
@aangelisc also quick question: when will there be a new azure-data-explorer-datasource plugin release?

aangelisc (approved):
LGTM, thanks @naizerjohn-ms

aangelisc:
> @aangelisc also quick question: when will there be a new azure-data-explorer-datasource plugin release for Grafana?

There are some other PRs to be merged and then a release will be cut 😊

PR #1196, opened 2025-02-19
Fixed query timeout formatting for timeout > 1m

nadav-dav:
Setting query timeout value that is larger than a minute caused the value that is sent to the server to have a trailing ")" in the end of it. This causes the timeout value to be ignored and fallback to the default value.

Notice the trailing comma in line 99 of the source file

### DIFF
--- a/pkg/azuredx/models/settings.go
+++ b/pkg/azuredx/models/settings.go
@@ -7,8 +7,9 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/grafana/azure-data-explorer-datasource/pkg/azuredx/helpers"
 	"github.com/grafana/grafana-plugin-sdk-go/backend"
+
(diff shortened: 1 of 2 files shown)


pkg/azuredx/models/settings.go
```diff
@@ -87,16 +87,13 @@ func formatTimeout(d time.Duration) (string, error) {
...
-	}
-	if d < time.Minute {
-		return fmt.Sprintf("00:00:%02.0f", d.Seconds()), nil
-	}
-	tMinutes := d.Truncate(time.Minute)
+	hours := d / time.Hour
+	d -= hours * time.Hour
+	minutes := d / time.Minute
+	d -= minutes * time.Minute
+	seconds := d / time.Second
```
ikonst, line 94:
```suggestion
	hours := int(d.Hours())
	minutes := int(d.Minutes()) % 60
	seconds := int(d.Seconds()) % 60
```

PR #1267, opened 2025-04-25
Remove openai for llm grafana lib

bossinc:
- update `openai` to `llm` in QueryHeader
- check if databases are in stateschema so null exception doesn't happen
- rename enabled to better reflect what its enabling


alyssabull (approved):
LGTM 🚀

PR #1291, opened 2025-05-27
Add support for comparison operators with `datetime` and `timespan` types

### LINKED ISSUE #457: Query builder: Unable to filter based on timespan or datetime ranges
**What happened**:

The builder quotes `timespan` or `datetime`s when comparing them in a filter. This syntax is only valid with the operators that compare exact values (e.g. `==` or `!=`) but it does not work with ranges:

Screenshot from 2022-08-25 16-55-15

To compare ranges, we need to explicitly convert them to the type, for example:

```
| where Duration > timespan('00:00:01')
```

**Screenshots**

**Environment**:

- Grafana version:
- Plugin version: 4.1.5
- OS Grafana is installed on:
- User OS & Browser:
- Others:

aangelisc:
This PR supports correctly casting values in where clauses when the type is `datetime` or `timespan`. It simply wraps the values in the correct type.

It should be noted that the value used in the operator must match exactly when using the `==` or `!=` operators whereas the `<`, `>`, `<=`, and `>=` operators are more lenient.

Fixes #457.


adamyeats (approved):

src/components/QueryEditor/VisualQueryEditor/utils/utils.ts
```diff
@@ -292,11 +297,15 @@ export const createOperator = (property: string, operator: string, value: any):
   };
 };
 
-export const valueToPropertyType = (value: any): QueryEditorPropertyType => {
+export const valueToPropertyType = (value: any, typeOverride?: QueryEditorPropertyType): QueryEditorPropertyType => {
   if (Array.isArray(value) && value.length > 0) {
```
beejeebus, line 301: does this need to be changed to also pass along the optional `typeOverride`?

PR #1301, opened 2025-06-09
Internationalisation: Add markup for translations

ashharrison90:
this looks bigger than it is! don't panic 😅 
what it does:
- adds translation markup
- adds the new i18n lint rule to ensure everything stays translated
- as a result of adding the lint rule, it also updates to eslint 9 and the new flat config format since that's the only way the new lint rule can be used
- as a result of updating to eslint 9, we also need to update to yarn 4 since otherwise we hit [a very annoying yarn bug similar to this](https://github.com/storybookjs/storybook/issues/22431#issuecomment-1630086092)
  - note this is from the storybook repo, but it's a generic yarn bug
  - this is what causes most of the changes - the lockfile format completely changed between yarn 1/yarn 2+
  
some things to note:
- version compatibility:
  - translations will work correctly in grafanaVersion >= 12.0.0 
  - will fallback to english in 10.2.0 <= grafanaVersion < 12.0.0 
  - will probably fail horribly for grafanaVersion < 10.2.0 - but that's ok, that version is unsupported and the minVersion is already 10.4.0 🥳 
- the crowdin github actions + corresponding translations for other languages will come in a later PR - they're not needed to maintain compatibility (selecting a different language will always fallback to english if it doesn't exist)

Fixes https://github.com/grafana/grafana/issues/105396


.yarn/install-state.gz
joshhunt, line 1: I don't think this file should be committed https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored

> `.yarn/install-state.gz` is an optimization file that you shouldn't ever have to commit. It simply stores the exact state of your project so that the next commands can boot without having to resolve your workspaces all over again.

PR #1338, opened 2025-06-30
Improve setting of initial variables and undefined type-checking

aangelisc:
Some variables do not have the correct initial values set. This can lead to some kind of race condition which can cause the editor to fail to load if a user is switching between data sources quickly.

I've improved the type-checking and I've also set initial values that are memoised and will update when the relevant dependency is changed.


adamyeats (commented):
LGTM, but needs tests ideally - in this case, however, modifications to the existing tests might be more appropriate

dreddy-work:
@aangelisc, When do we expect this to be released ?

adamyeats (approved):
Nice, LGTM! 👍

PR #1363, opened 2025-07-17
Support user-trusted endpoints

### LINKED ISSUE #864: Add support for user-specified trusted endpoints
Currently the trusted endpoints configuration only includes Azure endpoints for each cloud. As described in #849, this is unsuitable in cases where the clusters may sit behind a load-balancer for failover reasons - see [this](https://github.com/grafana/azure-data-explorer-datasource/pull/849#issuecomment-2044924818) comment for more details.

We should determine how to solve this whilst not worsening security in some way.

aangelisc:
Adds support for users to specify their own trusted endpoints. This can be useful in cases where multiple ADX clusters may be sitting behind a load-balancer or proxy but the Grafana server admin still wishes to restrict what endpoints the user can send requests to.

This feature must be enabled by setting the relevant flags in the Grafana config as below (where `user_trusted_endpoints` is a comma delimited list):

```ini
[plugin.grafana-azure-data-explorer-datasource]
enforce_trusted_endpoints = true
allow_user_trusted_endpoints = true
user_trusted_endpoints = https://first.endpoint.com,https://endpoint.second.com
```

Added a note in the README that this feature should be used with caution to avoid unintended sharing of auth tokens with 3rd parties.

Fixes #864

### DIFF
--- a/pkg/azuredx/client/httpclient.go
+++ b/pkg/azuredx/client/httpclient.go
@@ -81,6 +81,11 @@ func getAuthOpts(azureSettings *azsettings.AzureSettings, dsSettings *models.Dat
 		if err != nil {
 			return nil, err
 		}
+
+		if dsSettings.AllowUserTrustedEndpoints && len(dsSettings.UserTrustedEndpoints) > 0 {
+			endpoints = append(endpoints, dsSettings.UserTrustedEndpoints...)
+		}
+
 		err = authOpts.AllowedEndpoints(endpoints)
 		if err != nil {
 			return nil, err
--- a/pkg/azuredx/client/httpclient.go
+++ b/pkg/azuredx/client/httpclient.go
@@ -77,7 +77,7 @@ func getAuthOpts(azureSettings *azsettings.AzureSettings, dsSettings *models.Dat
 
 	// Enforce only trusted Azure Data Explorer endpoints if enabled
 	if userProvidedEndpoint && dsSettings.EnforceTrustedEndpoints {
-		endpoints, err := getAdxEndpoints(azureCloud)
+		endpoints, err := getAdxEndpoints(azureCloud, azureSettings)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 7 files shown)


pkg/azuredx/client/httpclient.go
```diff
@@ -81,6 +81,11 @@ func getAuthOpts(azureSettings *azsettings.AzureSettings, dsSettings *models.Dat
 		if err != nil {
 			return nil, err
 		}
+
+		if dsSettings.AllowUserTrustedEndpoints && len(dsSettings.UserTrustedEndpoints) > 0 {
+			endpoints = append(endpoints, dsSettings.UserTrustedEndpoints...)
```
JonCole, line 86: The method getAdxEndpoints has a limitation where it doesn't work for custom clouds provided through the [clouds_config setting](https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#clouds_config) because it throws an error if the cloud name is not known.  I think that means that this feature won't work for custom clouds.  I think we will need to re-work getAdxEndpoints so that it does not throw for an unknown clouds.

pkg/azuredx/models/settings.go
```diff
@@ -100,3 +115,11 @@ func envBoolOrDefault(key string, defaultValue bool) (bool, error) {
 		return value, nil
 	}
 }
+
+func envStringSliceOrDefault(key string, defaultValue []string) ([]string, error) {
+	if strValue := os.Getenv(key); strValue == "" {
+		return defaultValue, nil
+	} else {
+		return strings.Split(strValue, ","), nil
+	}
```
bossinc, line 124:
```suggestion
	return strings.Split(strValue, ","), nil
```

bossinc (approved):
Nice work!

bossinc (approved):
Nit go code formatting

pkg/azuredx/client/httpclient.go
JonCole: Looks good - thanks!

JonCole: @aangelisc safe to assume this will only work on 12.x bits?  Have you tested these changes on 11.x to make sure they are back compatible?

PR #1369, opened 2025-07-29
Wrap `module.ts` in `async` function to allow translations to be loaded before datasource init

adamyeats:
As top-level `await` appears to not be supported by the runtime of older versions of Grafana, this PR allows the intended order of execution in `module.ts` to remain the same, while keeping the intended export of `DataSourcePlugin`.


ashharrison90 (commented):
have you tested the behaviour when running against grafana@10.4.x with a different language? 🤔

ashharrison90:
i think this is also just working by accident... we've changed the return type of `plugin` to be a promise.

the only reason this is continuing to work is because we're already in a promise chain when we load the plugin [here](https://github.com/grafana/grafana/blob/main/public/app/features/plugins/pluginLoader.ts#L93-L97). but it means, for example, the `meta` won't get set correctly on the plugin 😬

PR #1370, opened 2025-07-29
Update workflow

aangelisc:
To ensure we're catching issues on all Grafana versions we support we should run the E2E tests all the way down to the minimum version we support. This change includes switching over to the default workflow and removing the redundant one.


bossinc (approved):

bossinc:
`/home/runner/work/azure-data-explorer-datasource/azure-data-explorer-datasource/src/plugin.json` its adding adx path twice. ohhhhhhhh CI

PR #1373, opened 2025-07-30
Prepare 6.2.0

bossinc:
Bump version and update changelog
Removing `initPluginTranslations` function. Using a root-level await isn't available in Grafana v10.4.
We are releasing 6.2.0 for Grafana v10.4 and we will release 7.0.0 with Grafana v11.0.0 requirement and the plugin translations.


aangelisc (commented):
Do we need to disable the translation workflows?

aangelisc (changes requested):
Blocking until we're sure this will work

ashharrison90 (changes requested):
i'm not sure this will work with 10.4.0 correctly unless we completely remove the markup. we also definitely need to remove the i18n workflows. 

i think the safest thing to do is:
- revert the 4 PRs i raised in reverse order https://github.com/grafana/azure-data-explorer-datasource/pulls?q=is%3Apr+author%3Aashharrison90+is%3Aclosed 
  - we particularly need to make sure the crowdin upload does not exist anymore 
- release 6.2.0
- reapply the PRs in normal order again
- release 7.0.0

PR #1460, opened 2025-10-01
Fix variable identification with Scenes

### LINKED ISSUE #1459: Multi-value template variables not correctly quoted
Multi-valued template variables are incorrectly quoted since the enablement of Scenes. This is due to some changes in the variable model.

See https://github.com/grafana/support-escalations/issues/18114 for more details.

aangelisc:
The enablement of scenes has removed the `id` property from template variables. In order to support correctly identifying template variable values I've swapped the logic to use `name` which seems to be consistent across the old and new implementations.

Needed for https://github.com/grafana/support-escalations/issues/18114

Fixes #1459

### DIFF
--- a/src/KustoExpressionParser.test.ts
+++ b/src/KustoExpressionParser.test.ts
@@ -853,6 +853,7 @@ describe('KustoExpressionParser', () => {
               text: 'usa',
               value: 'USA',
             },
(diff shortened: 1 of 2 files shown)


bossinc (approved):
nice!

PR #1466, opened 2025-10-02
Improved allowed endpoints support

### LINKED ISSUE #1429: Support wildcard in user_trusted_endpoints list
We have a customer that wants to specify user trusted endpoints with a simple wildcard format.

For example, this customer has multiple ADX endpoints in a format like "https://adx.<regionName>.example.com".  

The list of regions is expected to change over time (e.g. add new regions).  Rather than having to maintain/update the user_trusted_endpoints list to add new regions as they come online, the customer is asking if it would be possible to add some type of simple wildcard support such as "https://adx.*.example.com/".

aangelisc:
Bumping the `grafana-azure-sdk-go` to broaden the allowed endpoints support. Documentation has been updated accordingly.

Fixes #1429


adamyeats (approved):
🚀

PR #1566, opened 2025-12-31
Fix management commands failing due to DataSet_Table_0 naming

### LINKED ISSUE #1563: ADX .show management commands no longer return Table_0 – plugin fails to convert response to data frame
**What happened**:

Since yesterday, all Azure Data Explorer management/control commands (e.g. .show database details) stopped working in Grafana when using the grafana-azure-data-explorer-datasource plugin. Normal KQL queries (e.g. `print now()`) continue to work as expected.

Grafana returns the following error:

`no data as Table_0 table is missing from the the response`

ADX management/control command results should be correctly executed and converted into Grafana data frames, as they were previously.

- Configure an Azure Data Explorer datasource using the grafana-azure-data-explorer-datasource plugin.
- Open Explore or create a dashboard panel.
- Run a management command, for example: `.show database details`

Observe the error: no data as Table_0 table is missing from the response.

**Screenshots**

Plugin version:

Image

Grafana responce:

Image

**Environment**:

- Grafana version: 12.3.1
- Plugin version: 7.2.1
- OS Grafana is installed on: Azure App Service (Linux)
- User OS & Browser: macos/chrome

uristernik:
I have noticed that ~Dec 23, 2025, ADX management commands (e.g., .show commands) return tables named DataSet_Table_0 instead of Table_0. This caused the plugin to fail with: "no data as Table_0 table is missing from the response"

Fixes #1563

----
### Solution
Added getPrimaryResultTable() that gets the first table in the response regardless the name


pkg/azuredx/models/table.go
```diff
@@ -39,17 +39,35 @@ type Column struct {
...
+	return Table{}, fmt.Errorf("no data as %v table is missing from the response", name)
+}
+
+// getPrimaryResultTable returns the primary result table from the response.
+// It handles different naming conventions used by ADX for query vs management commands:
+// - Query responses use: Table_0, Table_1, etc.
+// - Management command responses use: DataSet_Table_0, DataSet_Table_1, etc.
+func (tr *TableResponse) getPrimaryResultTable() (Table, error) {
+	// Try standard query response format first
+	if table, err := tr.getTableByName("Table_0"); err == nil {
```
zivcaspi: The "TableName" field in a Kusto (Azure Data Explorer) response is non-contractual, and is subject to change either by a code change in the service itself or by the specifics of the query/control command being issued. I'm not sure why this code doesn't use the Kusto Go SDK, but if making a change the recommendation is to simply pick up the first table being sent back (I don't know if Grafana supports more than one primary result table in the response stream; if it does then this code doesn't parse it correctly anyway.)

PR #1678, opened 2026-03-10
Add key words to escape

bossinc:


bossinc:
Alternative approach to https://github.com/grafana/azure-data-explorer-datasource/pull/1677

PR #1749, opened 2026-05-21
remove reference to current user auth as experimental

katebrenner:
This PR removes all code that references current user auth as experimental, as this feature is no longer experimental


src/components/ConfigEditor/AzureCredentialsForm.tsx
```diff
@@ -158,29 +158,6 @@ export const AzureCredentialsForm: FunctionComponent<Props> = (props: Props) =>
           />
         </Field>
       )}
-      {credentials.authType === 'currentuser' && (
```
aangelisc, line 161: I think we may still want a warning about limitations?

katebrenner: good call, added back in

adamyeats (approved):

PR #1753, opened 2026-05-26
chore: set package manager

yesoreyeram:
Follow-up of https://github.com/grafana/azure-data-explorer-datasource/pull/1736


aangelisc (changes requested):
Can we upgrade npm in this PR as well please

aangelisc (approved):

PR #1782, opened 2026-06-30
Add Kusto endpoint for AzureUSGovernment

msoumar-ms:
When accessing Kusto databases in Azure US Government (Fairfax), the current scope is /.default. Add the Kusto endpoint for USGov to change this to kusto.kusto.usgovcloudapi.net/.default.


aangelisc (approved):
Thank you @msoumar-ms!

PR #1815, opened 2026-08-14
feat: per-request metadata resolution

aangelisc:
The original implementation of this feature relied on the default cluster URL. As we support users switching the cluster in the builder, we also need to support dynamic resolution of the cluster scope.

This change uses a new SDK property that allows us to specify a custom scope resolver. In the Kusto HTTP client, we will now resolve the cluster scope per request based on the cluster URL specified in the request.

The scopes will be cached up to 1h per-cluster. We use a `sync.Map` to ensure there is no read contention. Additionally, we use a `singleflight.Group` to collapse requests to the same cluster's metadata endpoint. Finally, each request has it's own context to ensure the cancellation of one does not cancel others. The metadata request context timeout is kept low (10s) to ensure it doesn't block for too long.

If the metadata response does not resolve, then the data sources' default cluster URL will be used. This may not work for all clusters, but should be okay for the majority of ADX clusters in the same cloud. This is non-negotiable.

I've also fixed a minor UI bug where an error could show when switching the cluster, before the schema request had actually resolved.


pkg/azuredx/client/metadata.go
```diff
@@ -75,3 +76,23 @@ func fetchAuthMetadata(ctx context.Context, httpClient *http.Client, clusterURL
...
+
+		scopes, err := cache.resolve(ctx, clusterUrl, func(fetchCtx context.Context) ([]string, error) {
+			metadata, err := fetchAuthMetadata(fetchCtx, metadataClient, clusterUrl)
+			if err != nil {
+				return nil, err
+			}
+			return []string{fmt.Sprintf("%s/.default", metadata.KustoServiceResourceID)}, nil
+		})
+		if err != nil {
+			backend.Logger.FromContext(ctx).Warn("failed to fetch auth metadata from cluster, falling back to default scopes", "cluster", clusterUrl, "error", err)
```
adamyeats, line 103: Consider distinguishing `ctx.Err()` from a real fetch failure before logging here. A routine panel unmount or dashboard refresh cancels the in-flight metadata fetch and hits this same warning as a genuine cluster outage, which will make normal usage look like a flaky endpoint.

pkg/azuredx/client/scope_cache.go
```diff
@@ -0,0 +1,85 @@
...
+			return scopes, nil
+		}
+
+		fetchCtx, cancel := context.WithTimeout(context.Background(), metadataTimeout)
+		defer cancel()
+
+		scopes, err := fetch(fetchCtx)
+		if err != nil {
+			return nil, err
+		}
```
adamyeats: It might be worth caching a failed fetch too (even briefly). Right now a metadata-endpoint outage means every non-overlapping query re-pays the full 10s `metadataTimeout`, where the old code resolved once at construction and reused the fallback for the datasource's lifetime.

src/components/QueryEditor/QueryEditor.tsx
```diff
@@ -48,7 +48,7 @@ export const QueryEditor: React.FC<Props> = (props) => {
 
   return (
     <>
-      {schema.error && (
+      {!schema.loading && schema.error && (
```
adamyeats, line 53: This guard fixes the error banner, but `schema.value` is still passed through unguarded a bit further down, at lines 76, 86 and 95 (`OpenAIEditor`, `RawQueryEditor`, `VisualQueryEditor`). Since `useAsync` spreads previous state on rerun, switching cluster/database keeps feeding the previous schema into those editors while the new one loads, so a user could build a query against tables that don't exist in the newly selected database.

pkg/azuredx/client/scope_cache.go
```diff
@@ -0,0 +1,85 @@
...
+
+		c.entries.Store(key, scopeCacheEntry{scopes: scopes, expiry: time.Now().Add(c.ttl)})
```
adamyeats: `entries` (the `sync.Map` declared above at line 27) has no eviction, so every distinct cluster host ever successfully resolved stays cached for the instance's lifetime. Worth confirming that's acceptable given `EnforceTrustedEndpoints` defaults to `false`, since the key space is effectively whatever `ClusterUri` a query specifies.

pkg/azuredx/client/httpclient.go
```diff
@@ -119,3 +117,23 @@ func getHttpClient(ctx context.Context, instanceSettings *backend.DataSourceInst
...
 	return httpClient, nil
 }
+
+// metadataTimeout bounds both the metadata HTTP client and the detached
+// context used to fetch cluster auth metadata, so a slow or unreachable
+// metadata endpoint falls back to default scopes promptly.
+const metadataTimeout = 10 * time.Second
+
+func newMetadataHTTPClient(ctx context.Context, instanceSettings *backend.DataSourceInstanceSettings) (*http.Client, error) {
+	clientOpts, err := instanceSettings.HTTPClientOptions(ctx)
```
adamyeats: Two changes compound here: the metadata timeout dropped from 30s to 10s, and the metadata client now inherits the datasource's proxy/TLS/basic-auth settings via the same options the query client uses. A cluster that's slow-but-reachable through a corporate proxy could now reliably miss this budget and fall back to the wrong scope for a full hour (`scopeCacheTTL`), where it previously cleared 30s fine.

pkg/azuredx/client/httpclient.go
```diff
@@ -27,15 +27,13 @@ func newHttpClientAzureCloud(ctx context.Context, instanceSettings *backend.Data
...
-			Timeout: 30 * time.Second,
-		},
-	})
+	metadataClient, err := newMetadataHTTPClient(ctx, instanceSettings)
 	if err != nil {
-		return nil, fmt.Errorf("error creating metadata http client: %w", err)
+		return nil, err
 	}
-	scopes, err := getAdxScopes(ctx, metadataClient, azureCloud, dsSettings.ClusterURL)
+	authOpts.SetScopeResolver(newScopeResolver(azureCloud, metadataClient))
```
adamyeats: nit: `authOpts.Scopes(scopes)` a few lines below (line 40) is unreachable today since the resolver's own fallback never surfaces an error to the SDK, but it's seeded from the datasource's default cluster. If the resolver above is ever changed to propagate real errors, this would silently apply the wrong cluster's audience.

pkg/azuredx/client/httpclient.go
```diff
@@ -119,3 +117,23 @@ func getHttpClient(ctx context.Context, instanceSettings *backend.DataSourceInst
 
 	return httpClient, nil
 }
+
+// metadataTimeout bounds both the metadata HTTP client and the detached
+// context used to fetch cluster auth metadata, so a slow or unreachable
+// metadata endpoint falls back to default scopes promptly.
+const metadataTimeout = 10 * time.Second
+
+func newMetadataHTTPClient(ctx context.Context, instanceSettings *backend.DataSourceInstanceSettings) (*http.Client, error) {
```
adamyeats: nit: `newMetadataHTTPClient` duplicates `getHttpClient`'s construction skeleton almost verbatim. Consider extracting the shared bit so the two clients can't quietly drift apart.

pkg/azuredx/client/scope_cache.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	ch := c.group.DoChan(key, func() (interface{}, error) {
+		// Another caller may have populated the cache between our miss above and
+		// acquiring this flight.
+		if scopes, ok := c.get(key); ok {
+			return scopes, nil
+		}
+
+		fetchCtx, cancel := context.WithTimeout(context.Background(), metadataTimeout)
```
adamyeats: nit: the detached fetch context is rooted in a bare `context.Background()`. Deriving it from something like `context.WithoutCancel(ctx)` would keep the metadata call traceable as a child of the triggering request.

pkg/azuredx/client/metadata.go
```diff
@@ -75,3 +76,23 @@ func fetchAuthMetadata(ctx context.Context, httpClient *http.Client, clusterURL
...
+	cache := newScopeCache(scopeCacheTTL)
+	return func(ctx context.Context, req *http.Request) ([]string, error) {
+		clusterUrl := fmt.Sprintf("%s://%s", req.URL.Scheme, req.URL.Host)
+
+		scopes, err := cache.resolve(ctx, clusterUrl, func(fetchCtx context.Context) ([]string, error) {
+			metadata, err := fetchAuthMetadata(fetchCtx, metadataClient, clusterUrl)
+			if err != nil {
+				return nil, err
+			}
+			return []string{fmt.Sprintf("%s/.default", metadata.KustoServiceResourceID)}, nil
```
adamyeats: nit: ADX only ever has one audience scope. Since `scopeCache` is package-private with a single call site, it might be simpler to cache a single `string` and wrap it into `[]string{scope}` once at the `ScopeResolver` boundary.

adamyeats (changes requested):
Comments inline, with extra attention on the caching/concurrency parts as requested.

pkg/azuredx/client/metadata.go
aangelisc: Done!

src/components/QueryEditor/QueryEditor.tsx
aangelisc: Good catch 👍

pkg/azuredx/client/httpclient.go
aangelisc: Done

pkg/azuredx/client/httpclient.go
aangelisc: Added a comment explaining the logic

pkg/azuredx/client/httpclient.go
aangelisc: Bumped timeout to 30s

pkg/azuredx/client/scope_cache.go
aangelisc: Added eviction when trying to get an expired entry, lmk if this is sufficient

adamyeats (approved):

PR #1816, opened 2026-08-18
VariableEditor: add hint about column names

ikonst:


ikonst:
On #7, users have rightfully noticed that while there's a mapping, there's nothing documenting the expected column names.

ikonst:
@iwysiu care to take a look?

